PR #6, opened 2020-07-21
Mocked search algorithm for prototype

Domene99:
Mocked some query calls as comments, and some articles. And manually extracted keywords. Counted the amount of indices between every article and a single query. I'd say this mocking is enough to see this as a viable solution.

### DIFF
--- a/prototyping/mocking/mockSearch.cpp
+++ b/prototyping/mocking/mockSearch.cpp
@@ -1,0 +1,48 @@
+#include <bits/stdc++.h>
+using namespace std;
+
+void search(unordered_set<string> words, vector<string> articles) {
+	unordered_map<string, int> incidences;
+
+	for (const auto & article : articles) {
+		string curr_word = "";
+		
+		for (const auto & letter : article) {
+			if (!isalpha(letter)) {
+				if (words.find(curr_word) != words.end()) {
+					cout << curr_word << "\n";
+					incidences[article]++;
+				}
+
+				curr_word = "";
+			} else {
+				curr_word += letter;
+			}
+		}
+		if (words.find(curr_word) != words.end()) {
+			incidences[article]++;
+		}
+	}
+
+	for (auto it = incidences.begin(); it != incidences.end(); it++) {
+		cout << "\t" << it->first << "\n\tcount: " << it->second << "\n\n";
+	}
+} 
+
+int main() {
+	vector<string> articles{"Los conductores y ocupantes de bicicletas, bicimotos o triciclos podran utilizar de preferencia casco de proteccion para su seguridad.",
+	       	"La autoridad municipal podra determinar la instalación de relojes estacionometros en la via publica, previo estudio de factibilidad elaborado por la Secretaria.",
+		"No se autorizara ningun exclusivo para taxis a una distancia menor de 300-trescientos metros a la redonda de otro sitio de taxis autorizados.",
+		"Las calcomania del refrendo del vehiculo deberan colocarse en el espacio establecido en la placa de circulacion."};
+
+	unordered_set<string> keywords_article_1 = {"obligatorio", "utilizar", "casco"}; // Es obligatorio el uso de casco
+	unordered_set<string> keywords_article_2 = {"via", "publica", "relojes"}; // Hay relojes en la via publica
+	unordered_set<string> keywords_article_3 = {"taxis", "exclusivo", "autorizados", "radio"}; // taxis exclusivo autorizados radio
+	unordered_set<string> keywords_article_4 = {"calcomania", "placa"}; // se cayo mi placa calcomania
+	vector<unordered_set<string>> queries { keywords_article_1, keywords_article_2, keywords_article_3, keywords_article_4 };
+	
+	for (int i = 0; i < queries.size(); i++) {
+		cout << "Query #" << i + 1 << ":\n";
+		search(queries[i], articles);
+	}
+}
--- a/prototyping/mocking/mockSearch.cpp
+++ b/prototyping/mocking/mockSearch.cpp
@@ -1,7 +1,14 @@
 #include <bits/stdc++.h>
 using namespace std;
 
-void search(unordered_set<string> words, vector<string> articles) {
+// search function
+// input: array of article text
+// input: set of words to search through
+// return: frequency map of articles and amount of incidences between 
+// 	   words from the text and words in the set
+// debugging output: every match between a word from an article and a
+// 		     word from the set is printed
+unordered_map<string, int> search(unordered_set<string> words, vector<string> articles) {
 	unordered_map<string, int> incidences;
 
 	for (const auto & article : articles) {
@@ -23,10 +30,7 @@ void search(unordered_set<string> words, vector<string> articles) {
 			incidences[article]++;
 		}
 	}
-
-	for (auto it = incidences.begin(); it != incidences.end(); it++) {
-		cout << "\t" << it->first << "\n\tcount: " << it->second << "\n\n";
-	}
+	return incidences;
 } 
 
 int main() {
@@ -43,6 +47,11 @@ int main() {
 	
 	for (int i = 0; i < queries.size(); i++) {
 		cout << "Query #" << i + 1 << ":\n";
-		search(queries[i], articles);
+		unordered_map<string, int> incidences = search(queries[i], articles);
+		
+		// Prints all articles which had a match for the specific query
+		for (auto it = incidences.begin(); it != incidences.end(); it++) {
+			cout << "\t" << it->first << "\n\tcount: " << it->second << "\n\n";
+		}
 	}
 }
--- a/prototyping/mocking/resultsOfMockSearch.txt
+++ b/prototyping/mocking/resultsOfMockSearch.txt
@@ -1,0 +1,26 @@
+Query #1:
+utilizar
+casco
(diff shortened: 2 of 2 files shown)


prototyping/mocking/mockSearch.cpp
```diff
@@ -0,0 +1,48 @@
+#include <bits/stdc++.h>
+using namespace std;
+
+void search(unordered_set<string> words, vector<string> articles) {
```
anniefu, line 4: I think defining the return here/printing it out is important for prototyping purposes. I see the ranking algorithm for what order you return articles in, and how you use NLP to help improve that, as the core functionality of the query engine.

prototyping/mocking/mockSearch.cpp
```diff
@@ -0,0 +1,48 @@
...
+	unordered_map<string, int> incidences;
+
+	for (const auto & article : articles) {
+		string curr_word = "";
+		
+		for (const auto & letter : article) {
```
anniefu, line 10: Have you and @jaimehisao had a chance to sync on the inputs/outputs of the parser?

It might be best to have the parser component to handle indexing the articles (creating a map from word to number of occurrences of a word in the article) during the initial parse of the article, since that never changes and can be done during the first pass.

jaimehisao: We haven't discussed that yet but it is definitely something that we need to do and define soon. If @Domene99 is okay with the parser doing the indexing I can make that happen.

Domene99: Yeah, I simply assumed that I'd receive an array of articles, which now is an array of jsons with the article full text as a value. Correct me if I'm wrong @jaimehisao

Domene99: And in the mvp we can discuss creating a frequency map for every article

prototyping/mocking/mockSearch.cpp
Domene99: Here I am just counting incidences. The nlp is sort of preprocessing for this algorithm. Full explanation is on my individual design doc.

Domene99:
> Disclaimer: I'm assuming you're mostly looking for conceptual feedback here, so I'm not leaving any style/data structure comments at this time.
> 
> Word count # seems like a reasonable prototype algorithm to me.
> 
> An interesting situation to think about is if the query is made up of a very common word and a very rare word. Like "parking at Google". In this case, which article ranks #1?
> 
> Article a: parking - 500 counts, Google - 1 count
> Article b: parking - 20 counts, Google - 5 counts

There can be keyword weights, Spencer suggested that a direct match could add more value than a synonym match. However, I'm not sure how viable it would be to weigh common v.s. uncommon words. Mainly because of the resources, even just finding a synonym api for spanish has been difficult.

prototyping/mocking/mockSearch.cpp
jaimehisao: Yes, the array of articles was the original plan. Sure, we can discuss that @Domene99

ssmall:
> > Disclaimer: I'm assuming you're mostly looking for conceptual feedback here, so I'm not leaving any style/data structure comments at this time.
> 
> > 
> 
> > Word count # seems like a reasonable prototype algorithm to me.
> 
> > 
> 
> > An interesting situation to think about is if the query is made up of a very common word and a very rare word. Like "parking at Google". In this case, which article ranks #1?
> 
> > 
> 
> > Article a: parking - 500 counts, Google - 1 count
> 
> > Article b: parking - 20 counts, Google - 5 counts
> 
> 
> 
> There can be keyword weights, Spencer suggested that a direct match could add more value than a synonym match. However, I'm not sure how viable it would be to weigh common v.s. uncommon words. Mainly because of the resources, even just finding a synonym api for spanish has been difficult.

I think this may be one huge drawback of your outlined approach. By breaking every query into keywords, you lose the ability to look for _exact matches_. I.e. if I search for "parking at Google" and there is one article that contains the exact phrase "parking at Google", that article is almost certainly more relevant than an article that contains the phrase "parking" 500 times and "Google" once.

Domene99:
> > > Disclaimer: I'm assuming you're mostly looking for conceptual feedback here, so I'm not leaving any style/data structure comments at this time.
> > 
> > 
> > > 
> > 
> > 
> > > Word count # seems like a reasonable prototype algorithm to me.
> > 
> > 
> > > 
> > 
> > 
> > > An interesting situation to think about is if the query is made up of a very common word and a very rare word. Like "parking at Google". In this case, which article ranks #1?
> > 
> > 
> > > 
> > 
> > 
> > > Article a: parking - 500 counts, Google - 1 count
> > 
> > 
> > > Article b: parking - 20 counts, Google - 5 counts
> > 
> > 
> > There can be keyword weights, Spencer suggested that a direct match could add more value than a synonym match. However, I'm not sure how viable it would be to weigh common v.s. uncommon words. Mainly because of the resources, even just finding a synonym api for spanish has been difficult.
> 
> I think this may be one huge drawback of your outlined approach. By breaking every query into keywords, you lose the ability to look for _exact matches_. I.e. if I search for "parking at Google" and there is one article that contains the exact phrase "parking at Google", that article is almost certainly more relevant than an article that contains the phrase "parking" 500 times and "Google" once.

I was thinking this could be a type of advanced setting. Sort of how google has quotation marks for direct results.

anniefu:
> > Disclaimer: I'm assuming you're mostly looking for conceptual feedback here, so I'm not leaving any style/data structure comments at this time.
> > Word count # seems like a reasonable prototype algorithm to me.
> > An interesting situation to think about is if the query is made up of a very common word and a very rare word. Like "parking at Google". In this case, which article ranks #1?
> > Article a: parking - 500 counts, Google - 1 count
> > Article b: parking - 20 counts, Google - 5 counts
> 
> There can be keyword weights, Spencer suggested that a direct match could add more value than a synonym match. However, I'm not sure how viable it would be to weigh common v.s. uncommon words. Mainly because of the resources, even just finding a synonym api for spanish has been difficult.

We've talked about this a bit offline now, but just to clarify. "common" vs "uncommon" would be with respect to all the texts that have been parsed (your apps "corpus" of documents), so you'd keep a total count of each word you've ever parsed, rather than the entire Spanish language.

anniefu:
> > > > Disclaimer: I'm assuming you're mostly looking for conceptual feedback here, so I'm not leaving any style/data structure comments at this time.
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > Word count # seems like a reasonable prototype algorithm to me.
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > An interesting situation to think about is if the query is made up of a very common word and a very rare word. Like "parking at Google". In this case, which article ranks #1?
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > Article a: parking - 500 counts, Google - 1 count
> > > 
> > > 
> > > > Article b: parking - 20 counts, Google - 5 counts
> > > 
> > > 
> > > There can be keyword weights, Spencer suggested that a direct match could add more value than a synonym match. However, I'm not sure how viable it would be to weigh common v.s. uncommon words. Mainly because of the resources, even just finding a synonym api for spanish has been difficult.
> > 
> > 
> > I think this may be one huge drawback of your outlined approach. By breaking every query into keywords, you lose the ability to look for _exact matches_. I.e. if I search for "parking at Google" and there is one article that contains the exact phrase "parking at Google", that article is almost certainly more relevant than an article that contains the phrase "parking" 500 times and "Google" once.
> 
> I was thinking this could be a type of advanced setting. Sort of how google has quotation marks for direct results.

I like the idea of making exact phrase matching an advanced setting. Another, simpler feature to think about supporting is exact single-word matching. Where you don't do any tokenizing or lemma extraction if the user requests an exact single-word match.

Domene99:
Seems everyone agrees with the PoC

> > > > > Disclaimer: I'm assuming you're mostly looking for conceptual feedback here, so I'm not leaving any style/data structure comments at this time.
> > > > 
> > > > 
> > > > > 
> > > > 
> > > > 
> > > > > Word count # seems like a reasonable prototype algorithm to me.
> > > > 
> > > > 
> > > > > 
> > > > 
> > > > 
> > > > > An interesting situation to think about is if the query is made up of a very common word and a very rare word. Like "parking at Google". In this case, which article ranks #1?
> > > > 
> > > > 
> > > > > 
> > > > 
> > > > 
> > > > > Article a: parking - 500 counts, Google - 1 count
> > > > 
> > > > 
> > > > > Article b: parking - 20 counts, Google - 5 counts
> > > > 
> > > > 
> > > > There can be keyword weights, Spencer suggested that a direct match could add more value than a synonym match. However, I'm not sure how viable it would be to weigh common v.s. uncommon words. Mainly because of the resources, even just finding a synonym api for spanish has been difficult.
> > > 
> > > 
> > > I think this may be one huge drawback of your outlined approach. By breaking every query into keywords, you lose the ability to look for _exact matches_. I.e. if I search for "parking at Google" and there is one article that contains the exact phrase "parking at Google", that article is almost certainly more relevant than an article that contains the phrase "parking" 500 times and "Google" once.
> > 
> > 
> > I was thinking this could be a type of advanced setting. Sort of how google has quotation marks for direct results.
> 
> I like the idea of making exact phrase matching an advanced setting. Another, simpler feature to think about supporting is exact single-word matching. Where you don't do any tokenizing or lemma extraction if the user requests an exact single-word match.

I'll add it to the design doc

Domene99:
Seems everyone agrees with the PoC so closing pr.

PR #7, opened 2020-07-21
Create CODEOWNERS

ssmall:
Add @ssmall and @anniefu as code owners for required approvals


anniefu:
Since there needs to be one mentor and one teammate review, let's add everyone to CODEOWNERS and bump required approvals up to two.

That way there's an escape hatch too if neither of the mentors or neither of the teammates are available.

PR #8, opened 2020-07-22
Initial setup of Web Application and Backend service.

### LINKED ISSUE #2: Frontend Prototype
A basic frontend that connects to a minimal backend

oscardavidrm:
Created an _app/_ folder encapsulating a React app instantiated with the ***create-react-app*** framework, and setup a _server/_ folder to setup an Apollo Server running on Node.js with an exposed GraphQL API.

Added as well basic configuration and deployment directives for App Engine, Cloud Build and Cloud Run services.

*** In order to run locally a .env file with necessary environment variables is necessary.

Closes #2


oscardavidrm:
> @googlebot I signed it!

jaimehisao:
Woah, thats a lot of new files!

ssmall:
What's the rationale for using GraphQL and Apollo for this project? What technical challenges are they addressing?

server/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+# A wildcard is used to ensure both package.json AND package-lock.json are copied.
+# Copying this separately prevents re-running npm install on every code change.
+COPY package*.json ./
+
+RUN yarn install --production
+
+# Copy local code to the container image.
+COPY . ./
+
+CMD [ "yarn", "start" ]
```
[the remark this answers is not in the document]
oscardavidrm: Yes!

This Dockerfile is consumed by Cloud Run to understand how to deploy the Docker container and what exactly to do afterwards.

So basically what this does, is copy all the server files (including dependencies) within the container and runs an instance of the server exposed to a default $PORT env variable (8080) set by Cloud Run, serving as an exposed endpoint.

[Here](https://graphql-mvnta4v2hq-uc.a.run.app/) I can show you the example running in my personal account.

app/src/App.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { Redirect, Route, Switch } from 'react-router-dom';
+import Loadable from 'react-loadable';
+import TopBarProgress from 'react-topbar-progress-indicator';
+import Layout from 'components/layout';
+
+/* webpackChunkName: "SearchEngine" */
+const SearchEngine = Loadable({
```
Domene99, line 8: Is this really necessary? I would guess that importing your own search engine component would be exactly the same as all other components

app/src/views/search-engine/index.js
```diff
@@ -0,0 +1,31 @@
+import React, { useState, useEffect } from 'react';
+import { useQuery } from '@apollo/client';
+import { useDebounce } from 'use-lodash-debounce';
+import { Container } from '@material-ui/core';
+import SearchBar from './components/search-bar';
+import ArticlesList from './components/articles-list';
+import { GET_ARTICLES } from './graphql/queries';
+
+const SearchEngine = () => {
```
Domene99, line 9: Seeing how this is the main component of the application some documentation would be good

app/src/views/search-engine/index.js
```diff
@@ -0,0 +1,31 @@
...
+  const [articles, setArticles] = useState([]);
+  const [search, setSearch] = useState('');
+  const debouncedSearch = useDebounce(search, 1000);
+  const getArticlesQuery = useQuery(GET_ARTICLES, { variables: { search: debouncedSearch } });
+
+  useEffect(() => {
+    const { data } = getArticlesQuery;
+
+    if (data?.articles) {
```
Domene99, line 18: Can you please explain what this operation does. If it's something not that known a single short comment would be nice. If it's something common then just answer my comment and leave it as is

server/graphql/resolvers/article/queries.js
```diff
@@ -0,0 +1,94 @@
+const articleQueries = {
+  articles: (_, { search }) => {
+    console.log(search);
```
Domene99, line 3: nit: remove

server/graphql/resolvers/article/queries.js
```diff
@@ -0,0 +1,94 @@
...
+    const articles = [
+      {
+        id: 1,
+        number: 1,
+        content:
+          'Esta prohibido manejar en las siguientes condiciones:\n\t 1. Teniendo los faros rotos\n\t 2. Teniendo las llantas ponchadas\nCualquiera de estos casos no será tolerado!',
+        keywords: [
```
Domene99, line 10: I wouldn't include this just yet, until we define it.

server/graphql/resolvers/article/queries.js
```diff
@@ -0,0 +1,94 @@
...
+          'Ipsum',
+          'Ipsum',
+          'Ipsum',
+          'Ipsum',
+          'Ipsum',
+          'Ipsum',
+          'Ipsum',
+          'Ipsum',
+        ],
+        minutesToRead: 5,
```
Domene99, line 22: Same goes for this, but this value doesn't affect the fundamental design so I'd say its alright if you leave it

server/Dockerfile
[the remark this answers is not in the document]
jaimehisao: nice! I like the Docker implementation! It would be nice to containerize all of our apps and have them in the repo too

app/src/views/search-engine/index.js
jaimehisao: +1

app/src/views/search-engine/components/articles-list/styles.js
```diff
@@ -0,0 +1,19 @@
...
+  cardContent: {
+    whiteSpace: 'pre-wrap',
+    marginBottom: 15,
+  },
+  keywords: {
+    width: '75%',
+  },
+}));
+
+export default styles;
```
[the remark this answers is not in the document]
oscardavidrm: Sure, will do.

This is the [@material-ui/core/styles](https://material-ui.com/styles/api/#makestyles-styles-options-hook) interface to encapsulate customized CSS styling, transitions, etc. So basically this is the ***Material UI*** way to add CSS (or at least one of the ways).

Now, the reason you see various `styles.js` files spread across the folders, its to down scope the styles to the React components they should be affecting. CSS has an inbuilt algorithm known as [cascade](https://developer.mozilla.org/en-US/docs/Web/CSS/Cascade), that basically determines which components get styled on the tree dependency, depending on where does the styles where declared. This algorithm has an exposed issue of potential conflicting rules that may have weird effects on the end result.

Mainly that's the reason I personally believe down scoping your CSS style to the same level you're gonna be using them drastically reduces conflicts and potential hard-tracking bugs.

app/src/App.js
oscardavidrm: This is known as Code [splitting/chunking](https://medium.com/fabhotels/code-splitting-and-chunking-frontend-apps-with-webpack-a15622388df) based on components to serve them on-demand. Although the current size of the application and files to be downloaded does not require code splitting, it is a technique to improve load time performance since we're only downloading the essential components that the user needs at any given section instead of downloading a monolithic bundle at once.

app/src/views/search-engine/index.js
oscardavidrm: These are known as [React Hooks](https://reactjs.org/docs/hooks-state.html#whats-a-hook). It was a feature added on recent iterations of React. It's basically the newest way to handle state within your components in an event-driven way.

This Hook is just querying the Apollo Server to fetch the articles and storing them in local state to display them.

server/graphql/resolvers/article/queries.js
oscardavidrm: Yeah, these are just mock up responses to design the webapp. The idea is that this is the generated result that should be returned and displayed to the client.

app/src/views/search-engine/index.js
oscardavidrm: What do you think is relevant information to include into the component?

PR #9, opened 2020-07-23
Basic Parser Functionality

jaimehisao:
Added basic parser without table functionality, that is in progress but will probably take more time to develop.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,138 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.nox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*.cover
+*.py,cover
+.hypothesis/
+.pytest_cache/
+cover/
+
+# Translations
+*.mo
+*.pot
+
+# Django stuff:
+*.log
+local_settings.py
+db.sqlite3
+db.sqlite3-journal
+
+# Flask stuff:
+instance/
+.webassets-cache
+
+# Scrapy stuff:
+.scrapy
+
+# Sphinx documentation
+docs/_build/
+
+# PyBuilder
+.pybuilder/
+target/
+
+# Jupyter Notebook
+.ipynb_checkpoints
+
+# IPython
+profile_default/
+ipython_config.py
+
+# pyenv
+#   For a library or package, you might want to ignore these files since the code is
+#   intended to run in multiple environments; otherwise, check them in:
+# .python-version
+
+# pipenv
+#   According to pypa/pipenv#598, it is recommended to include Pipfile.lock in version control.
+#   However, in case of collaboration, if having platform-specific dependencies or dependencies
+#   having no cross-platform support, pipenv may install dependencies that don't work, or not
+#   install all needed dependencies.
+#Pipfile.lock
+
+# PEP 582; used by e.g. github.com/David-OConnor/pyflow
+__pypackages__/
+
+# Celery stuff
+celerybeat-schedule
+celerybeat.pid
+
+# SageMath parsed files
+*.sage.py
+
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+
+# Spyder project settings
+.spyderproject
+.spyproject
+
+# Rope project settings
+.ropeproject
+
+# mkdocs documentation
+/site
+
+# mypy
+.mypy_cache/
+.dmypy.json
+dmypy.json
+
+# Pyre type checker
+.pyre/
+
+# pytype static type analyzer
+.pytype/
+
+# Cython debug symbols
+cython_debug/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.vscode
+./vscode
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+major-tom-venv
 .vscode
 ./vscode
 # Byte-compiled / optimized / DLL files
--- a/src/parser/parser.py
+++ b/src/parser/parser.py
@@ -51,10 +51,11 @@ def count_articles(pdf_text):
 
 
 # Acts as main
-with open("regs.pdf", "rb") as f:
-    doc = slate.PDF(f)
-    final_text = ""
-    for page in doc:
-        final_text += page
-    final_text = final_text.strip().lower().split()
-    count_articles(final_text)
+def parse():
+    with open("regs.pdf", "rb") as f:
+        doc = slate.PDF(f)
+        final_text = ""
+        for page in doc:
+            final_text += page
+        final_text = final_text.strip().lower().split()
+        count_articles(final_text)
--- a/src/parser/parser.py
+++ b/src/parser/parser.py
@@ -1,7 +1,9 @@
-import slate
-
-
+"""Parses a PDF document.
+"""
 import logging
+
+
+import slate
 
 
 logging.basicConfig(level=logging.INFO)
@@ -32,9 +34,9 @@ def count_articles(pdf_text):
     i = 0
     while i < len(pdf_text) - 1:
         if (pdf_text[i] == "artículo" or pdf_text[i] == "articulo") and (
-            pdf_text[i + 1] == str(article_count) + ".-"
-            or pdf_text[i + 1] == str(article_count) + "-"
-            or pdf_text[i + 1] == str(article_count) + "."
+                pdf_text[i + 1] == str(article_count) + ".-"
+                or pdf_text[i + 1] == str(article_count) + "-"
+                or pdf_text[i + 1] == str(article_count) + "."
         ):
             logging.info("Article #" + str(article_count) + " recognized!")
             articles.append(Article(article_count, article_text))
@@ -51,10 +53,12 @@ def count_articles(pdf_text):
 
 # Acts as main
 def parse():
-    with open("regs.pdf", "rb") as f:
-        doc = slate.PDF(f)
+    """Parses a PDF document
+    """
+    with open("regs.pdf", "rb") as pdf_file:
+        doc = slate.PDF(pdf_file)
         final_text = ""
         for page in doc:
-            final_text += page
+            final_text.join(page)
         final_text = final_text.strip().lower().split()
         count_articles(final_text)
--- a/src/parser/parser.py
+++ b/src/parser/parser.py
@@ -9,12 +9,24 @@ logging.basicConfig(level=logging.INFO)
 
 
 class Article:
+    """Class for storing articles.
+    """
+
     def __init__(self, number, text):
         self.number = number
         self.text = text
 
 
 def count_articles(pdf_text):
+    """Identifies articles and returns a list of Article objects.
+
+    Args:
+        pdf_text (string): contains the PDF text where articles 
+        are to be identified.
+
+    Returns:
+        list: article objects
+    """
     articles = []
     article_text = ""
     article_count = 1
@@ -35,18 +47,14 @@ def count_articles(pdf_text):
         i += 1
     for article in articles:
         logging.info("Article: " + str(article.number - 1) + " Text: " + article.text)
+    return articles
 
 
+# Acts as main
 with open("regs.pdf", "rb") as f:
     doc = slate.PDF(f)
-
     final_text = ""
     for page in doc:
-        why = page
         final_text += page
     final_text = final_text.strip().lower().split()
-    # final_text_pass = final_text
-    # final_text = numpy.array(final_text)
-    # numpy.savetxt("foo.txt", final_text, fmt="%s")
-    # print(final_text)
     count_articles(final_text)
--- a/src/parser/parser.py
+++ b/src/parser/parser.py
@@ -3,7 +3,6 @@ import slate
 
 import logging
 
-# import numpy
 
 logging.basicConfig(level=logging.INFO)
 
@@ -46,7 +45,7 @@ def count_articles(pdf_text):
             article_text += " " + pdf_text[i]
         i += 1
     for article in articles:
-        logging.info("Article: " + str(article.number - 1) + " Text: " + article.text)
+        logging.debug("Article: " + str(article.number - 1) + " Text: " + article.text)
     return articles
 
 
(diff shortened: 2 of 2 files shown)


src/parser/parser.py
```diff
@@ -0,0 +1,60 @@
...
+        are to be identified.
+
+    Returns:
+        list: article objects
+    """
+    articles = []
+    article_text = ""
+    article_count = 1
+    i = 0
+    while i < len(pdf_text) - 1:
```
Domene99, line 34: For enumerate might make this loop much cleaner. Enumerating gives both the index you are currently in and the value. https://treyhunner.com/2016/04/how-to-loop-with-indexes-in-python/#enumerate

src/parser/parser.py
```diff
@@ -0,0 +1,60 @@
...
+        if (pdf_text[i] == "artículo" or pdf_text[i] == "articulo") and (
+            pdf_text[i + 1] == str(article_count) + ".-"
+            or pdf_text[i + 1] == str(article_count) + "-"
+            or pdf_text[i + 1] == str(article_count) + "."
```
Domene99, line 38: You did the testing but please do tell me, is "articulo 7" not good enough. I know there might be multiple instances of that string but you do have article_count. Do you still need to check .-, -, and .?

src/parser/parser.py
```diff
@@ -0,0 +1,60 @@
...
+            logging.info("Article #" + str(article_count) + " recognized!")
+            articles.append(Article(article_count, article_text))
+            article_text = ""
+            article_count += 1
+            i += 1
+        else:
+            article_text += " " + pdf_text[i]
+        i += 1
+    for article in articles:
+        logging.info("Article: " + str(article.number - 1) + " Text: " + article.text)
```
Domene99, line 49: nit: make this a function in the article class, after all you are using only article attributes

src/parser/parser.py
```diff
@@ -0,0 +1,60 @@
...
+    article_count = 1
+    i = 0
+    while i < len(pdf_text) - 1:
+        if (pdf_text[i] == "artículo" or pdf_text[i] == "articulo") and (
+            pdf_text[i + 1] == str(article_count) + ".-"
+            or pdf_text[i + 1] == str(article_count) + "-"
+            or pdf_text[i + 1] == str(article_count) + "."
+        ):
+            logging.info("Article #" + str(article_count) + " recognized!")
+            articles.append(Article(article_count, article_text))
```
Domene99, line 41: Nice!

src/parser/parser.py
```diff
@@ -0,0 +1,60 @@
...
+
+    Returns:
+        list: article objects
+    """
+    articles = []
+    article_text = ""
+    article_count = 1
+    i = 0
+    while i < len(pdf_text) - 1:
+        if (pdf_text[i] == "artículo" or pdf_text[i] == "articulo") and (
```
Domene99, line 35: So won't titles and anything that goes before "articulo 1.-" be categorized as an article?

jaimehisao: no

src/parser/parser.py
```diff
@@ -0,0 +1,60 @@
+import slate
+
+
+import logging
+
+# import numpy
```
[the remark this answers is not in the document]
jaimehisao: What do you mean? the imports are of different groups, that´s why they are separated. Logging is part of system libraries and slate is imported.

src/parser/parser.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+# Acts as main
+with open("regs.pdf", "rb") as f:
+    doc = slate.PDF(f)
+    final_text = ""
+    for page in doc:
+        final_text += page
+    final_text = final_text.strip().lower().split()
+    count_articles(final_text)
```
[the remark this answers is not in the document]
jaimehisao: No need! The with statement automatically closes the file.

src/parser/parser.py
```diff
@@ -0,0 +1,60 @@
...
+            i += 1
+        else:
+            article_text += " " + pdf_text[i]
+        i += 1
+    for article in articles:
+        logging.info("Article: " + str(article.number - 1) + " Text: " + article.text)
+    return articles
+
+
+# Acts as main
```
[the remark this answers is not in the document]
jaimehisao: changed to not be a main

src/parser/parser.py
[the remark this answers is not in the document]
Domene99: I meant more of the spacing and the commented numpy

src/parser/parser.py
jaimehisao: usually it has the punctuation in the same token. will optimize later

src/parser/parser.py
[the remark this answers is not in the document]
jaimehisao: Oh! On Python you leave a whitespace between the # and the text

src/parser/parser.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+import logging
+
+# import numpy
+
+logging.basicConfig(level=logging.INFO)
+
+
+class Article:
```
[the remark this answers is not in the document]
jaimehisao: Can you link the specific style guide reference?

src/parser/parser.py
jaimehisao: While it is much cleaner, the behavior is not the same. I can't add to the variable `i` from inside the loop when using a for, thus I use a `while` loop.

src/parser/parser.py
```diff
@@ -0,0 +1,60 @@
...
+    return articles
+
+
+# Acts as main
+with open("regs.pdf", "rb") as f:
+    doc = slate.PDF(f)
+    final_text = ""
+    for page in doc:
+        final_text += page
+    final_text = final_text.strip().lower().split()
```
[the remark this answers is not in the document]
jaimehisao: those don't appear ;)
They dissapeared when I changed from `Py2PDF` to the `slate`Python library.

src/parser/parser.py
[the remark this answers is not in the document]
jaimehisao: I removed the numpy too

src/parser/parser.py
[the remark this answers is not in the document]
Domene99: Does the link not work? It sends you directly to the inner classes section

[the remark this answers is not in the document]
jaimehisao: It works, but it is not related. An Inner class is when you define a class inside a class...

src/parser/parser.py
Domene99: You're right. Though, sometimes a continue statement can serve a similar purpose

jaimehisao: Yes but a continue does not work in this case. I'm adding a one to `i` in order to skip adding the article number to the text.

src/parser/parser.py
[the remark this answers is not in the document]
Domene99: Yeah you're right.

src/parser/parser.py
```diff
@@ -0,0 +1,64 @@
...
+logging.basicConfig(level=logging.INFO)
+
+
+class Article:
+    """Class for storing articles.
+    """
+
+    def __init__(self, number, text):
+        self.number = number
+        self.text = text
```
oscardavidrm, line 18: This by no means is your responsibility, but I think we should align our attribute names to match throughout the entire application just to reduce friction in the convention.

I highlight this because in the web app, I am calling the body of every article `content`, and I believe that having different names on the different scripts and parts of the program, may be confusing to any external.

src/parser/parser.py
```diff
@@ -0,0 +1,64 @@
+"""Parses a PDF document.
```
oscardavidrm, line 1: What is the best convention in file naming? For classes should you capitalize the file, as in `Parser.py`, or is it different than other object-oriented programming languages?

I am using C++ as a benchmark, so it's very likely that it applies different in Python. But Idk.

src/parser/parser.py
```diff
@@ -0,0 +1,64 @@
...
+            i += 1
+        else:
+            article_text += " " + pdf_text[i]
+        i += 1
+    for article in articles:
+        logging.debug("Article: " + str(article.number - 1) + " Text: " + article.text)
+    return articles
+
+
+# Acts as main
```
oscardavidrm, line 54: I'm guessing all of this pattern matching is aimed specifically at the document we're parsing right? Like looking for either **artículo** or **articulo**.

I'm pretty sure that each document is gonna have its own patterns and is gonna be presented different, so I guess this makes sense. Although this is definitely out of the scope of what we're aiming for, I think it'd be good to do some brainstorming to think of ways to come up with a general algorithm that does a first clean up.

Imagine having to rewrite this every time we want to add a new geographical zone. 🥵

src/parser/parser.py
jaimehisao: module_name, package_name, ClassName, method_name, ExceptionName, function_name, GLOBAL_CONSTANT_NAME, global_var_name, instance_var_name, function_parameter_name, local_var_name.

src/parser/parser.py
jaimehisao: Yes, it is specific to our document. I understand that it would be nice to parse many documents with ease, but I see that as post MVP. Right now I see our focus to be to parse this document accurately.

src/parser/parser.py
jaimehisao: Yeah we should

src/parser/parser.py
Domene99, line 1: Maybe a little more descriptive comment?

src/parser/parser.py
```diff
@@ -0,0 +1,64 @@
...
+class Article:
+    """Class for storing articles.
+    """
+
+    def __init__(self, number, text):
+        self.number = number
+        self.text = text
+
+
+def count_articles(pdf_text):
```
Domene99, line 21: Why is it called count articles? I think this name doesn't fit

src/parser/parser.py
Domene99, line 54: nit: remove

src/parser/parser.py
```diff
@@ -0,0 +1,64 @@
...
+        else:
+            article_text += " " + pdf_text[i]
+        i += 1
+    for article in articles:
+        logging.debug("Article: " + str(article.number - 1) + " Text: " + article.text)
+    return articles
+
+
+# Acts as main
+def parse():
```
anniefu, line 55: This should take in the filename as an arg, now that it's not main.

src/parser/parser.py
```diff
@@ -0,0 +1,64 @@
...
+            logging.info("Article #" + str(article_count) + " recognized!")
+            articles.append(Article(article_count, article_text))
+            article_text = ""
+            article_count += 1
+            i += 1
+        else:
+            article_text += " " + pdf_text[i]
+        i += 1
+    for article in articles:
+        logging.debug("Article: " + str(article.number - 1) + " Text: " + article.text)
```
anniefu, line 50: Rather than doing another loop, probably just log this in the previous loop.

src/parser/parser.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __init__(self, number, text):
+        self.number = number
+        self.text = text
+
+
+def count_articles(pdf_text):
+    """Identifies articles and returns a list of Article objects.
+
+    Args:
```
anniefu, line 24: What counts as an Article? There's more than one article per PDF?

src/parser/parser.py
```diff
@@ -0,0 +1,64 @@
...
+                or pdf_text[i + 1] == str(article_count) + "."
+        ):
+            logging.info("Article #" + str(article_count) + " recognized!")
+            articles.append(Article(article_count, article_text))
+            article_text = ""
+            article_count += 1
+            i += 1
+        else:
+            article_text += " " + pdf_text[i]
+        i += 1
```
anniefu, line 48: It seems counterproductive to do a .split() on the entire PDF in `parse()`, but then rebuild the string here. 

I'd recommend using substring and finding the indices of your "article-separators" (i.e. "1.- "2-", "3.", etc). Then splitting the PDF text off the indices, instead of rebuilding the PDF text word-by-word.

src/parser/parser.py
```diff
@@ -0,0 +1,64 @@
...
+    Returns:
+        list: article objects
+    """
+    articles = []
+    article_text = ""
+    article_count = 1
+    i = 0
+    while i < len(pdf_text) - 1:
+        if (pdf_text[i] == "artículo" or pdf_text[i] == "articulo") and (
+                pdf_text[i + 1] == str(article_count) + ".-"
```
anniefu, line 37: This is a good opportunity to use regular expressions! https://www.w3schools.com/python/python_regex.asp

And also string formatting: https://pyformat.info/

If you're not familiar with regular expressions, your three `or`s and the if statement could be combined to something like: 
`re.split("art(i|e)culo {0}(-|.)-*".format(str(article_count))` I don't have an accented i on my keyboard for articulo, so I replaced it with an "e" for this example. 

Here's a rough explanation: https://regex101.com/r/N6DsXl/1/

(You just want full matches, you don't care about group matches in this case).

.gitignore
```diff
@@ -0,0 +1,141 @@
+major-tom-venv
```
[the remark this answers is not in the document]
jaimehisao: I used GitHub's recommended `.gitignore` for [Python](https://github.com/github/gitignore/blob/master/Python.gitignore) Still, I've gone ahead and removed most of the un-necessary stuff from there.

src/parser/parser.py
jaimehisao: yup!

src/parser/parser.py
jaimehisao: Went ahead and added the specifications to the comments

src/parser/parser.py
jaimehisao: Yes, I thought about using regex and @Domene99 suggested using it. I'm on it.

Domene99: With regex you can even skip the whole while loop!

src/parser/parser.py
```diff
@@ -0,0 +1,64 @@
+"""Parses a PDF document.
+"""
+import logging
+
+
+import slate
```
ssmall, line 6: Since this library must be installed in order for the parser to work, this PR should contain one of two things:
1) instructions for manually installing dependencies
2) encapsulation of dependencies using something like [pipenv][], [virtualenv][], or a [Dockerfile][]

[pipenv]: https://pipenv-fork.readthedocs.io/en/latest/
[virtualenv]: https://packaging.python.org/key_projects/#virtualenv
[Dockerfile]: https://runnable.com/docker/python/dockerize-your-python-application

src/parser/parser.py
ssmall, line 1: What version of Python is this?

src/parser/parser.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+logging.basicConfig(level=logging.INFO)
+
+
+class Article:
```
ssmall, line 12: Will `Article` have any methods defined on it, or is it just a container for fields? If it does not have any methods, consider using a [namedtuple][] instead.

[namedtuple]: https://docs.python.org/3.3/library/collections.html?highlight=namedtuple#collections.namedtuple

src/parser/parser.py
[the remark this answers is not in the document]
ssmall: It would be good to still include something like 
```python
if __name__ == '__main__':
    parse()
```

so that your prototype can be run directly from the command line.

src/parser/parser.py
jaimehisao: I'm using virtualenv

src/parser/parser.py
```diff
@@ -0,0 +1,64 @@
...
+
+    Returns:
+        list: article objects
+    """
+    articles = []
+    article_text = ""
+    article_count = 1
+    i = 0
+    while i < len(pdf_text) - 1:
+        if (pdf_text[i] == "artículo" or pdf_text[i] == "articulo") and (
```
oscardavidrm, line 36: LGTM!

Are you still planning on implementing regex in this part? Or will you leave that for another iteration?

jaimehisao: Yes! but for the next PR, which will be more complete

PR #10, opened 2020-07-24
Search algorithm first iteration

Domene99:
Mocked the article values from the db. Wrote 3 basic tests. Made Search algorithm it's own class for added functionality for future iterations.

Closing #3


prototyping/testing/searchTest.py
```diff
@@ -0,0 +1,54 @@
...
+                      "copa", "suelo", "pezuna", "vaso", "cabeza"]
+
+keywords_multiple = ["taxi", "tener", "calcomania", "placa"] # Mi taxi no tiene estampa en la placa
+synonyms_multiple = ["taximetro",
+                     "haber", "poseer", "detentar", "sujetar", "coger",
+                     "copia", "reproduccion", "estampa", "adhesivo",
+                     "lamina", "plancha", "chapa", "hoja", "pelicula"]
+
+
+def test_same_weights_for_all():
```
oscardavidrm, line 41: Although the testing functions do have a descriptive name of what they do, probably a small comment in plain english of what they aim to test would be nice.

prototyping/MockArticle.py
```diff
@@ -0,0 +1,8 @@
+class Article:
```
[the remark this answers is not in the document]
Domene99: When in doubt just check google's style guide,  https://google.github.io/styleguide/pyguide.html#3163-file-naming. Afaik, there's no problem with having different file and class name. 
I, however did this out of pure convinience, since it seems that everyone is individually making their own article I named it this way to avoid merge conflicts in the future.

prototyping/Search.py
```diff
@@ -0,0 +1,50 @@
+from MockArticle import Article
```
jaimehisao, line 1: [nit] Consider importing `MockArticle` as a whole and then use `MockArticle.Article`

prototyping/testing/searchTest.py
```diff
@@ -0,0 +1,63 @@
+import os
+import sys
+import inspect
+
+current_dir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
+parent_dir = os.path.dirname(current_dir)
+sys.path.insert(0, parent_dir) 
+from Search import Search
+from MockArticle import Article
```
jaimehisao, line 9: [nit} import the whole modules.

prototyping/testing/searchTest.py
```diff
@@ -0,0 +1,63 @@
+import os
```
jaimehisao, line 1: Very nice!

prototyping/Search.py
Domene99: I'm sorry but why? I myself think it looks better this way. As far as the style guide goes it says either way is okay https://google.github.io/styleguide/pyguide.html#224-decision

Domene99:
> Very nice algorithm! Looking closer to an MVP. It would be nice to standardize our `Article` class, especially us that we are using Python and our work will intersect a lot!

Well let's just push one of the article classes into the main, what do you think?

prototyping/Search.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        self.keywords = keywords
+        self.synonyms = synonyms
+        self.articles = articles
+        self.keywords_weight = keywords_weight
+        self.synonyms_weight = synonyms_weight
+
+    def score_articles(self):
+        """
+        Counts the number of indices between article words, keywords, and synonyms
```
anniefu, line 30: [nit] *incidences

prototyping/MockArticle.py
```diff
@@ -0,0 +1,8 @@
...
+
+    def __init__(self, number, content, keywords):
```
anniefu, line 3: Could you add a comment of what you expect these inputs to be too? I can figure it out from the code in Search.py, but it is not the most convenient way to figure it out. 

Also it will help you and @jaimehisao coordinate here if expectations are explicitly stated.

prototyping/Search.py
```diff
@@ -0,0 +1,50 @@
...
+            score = 0
+            article_frequency = article.keywords
+            for keyword in self.keywords:
+                score += article_frequency.get(keyword, 0) * self.keywords_weight
+
+            for synonym in self.synonyms:
+                score += article_frequency.get(synonym, 0) * self.synonyms_weight
+
+            if score != 0:
+                score_per_article[article] = score
```
anniefu, line 48: Seems more optimal to key this off article ID/number vs article itself. Not sure how python handles hashing classes, but I assume the entire article.content gets hashed if you do this.

prototyping/MockArticle.py
```diff
@@ -0,0 +1,8 @@
...
+        self.number = number
```
anniefu, line 4: assuming this is the same as an ID, maybe self.id?

prototyping/Search.py
```diff
@@ -0,0 +1,50 @@
+from MockArticle import Article
+
+class Search:
+    """
```
anniefu, line 4: Hmm, with this set up, it looks like you have to pass all existing articles with every query, which seems a bit inefficient. Maybe this should be a bit more multipurpose, like a SearchEngine class?

For a SearchEngine class, I would expect more like:
```python
def __init__(self, articles, settings):
   ...

def search(self, query):
   ...
```

prototyping/Search.py
```diff
@@ -0,0 +1,50 @@
...
+    Attributes:
+        keywords: A list of keywords
+        synonyms: A list of synonyms for the keywords
+        articles: A list of articles
+        keywords_weight: The weight that a keyword has on the scorig algorithm
+            Optional: default argument as 1
+        synonyms_weight: The weight that a synonym has on the scoring algorithm
+            Optional: default argument as 1
+    """
+    def __init__(self, keywords, synonyms, articles, keywords_weight=1, synonyms_weight=1):
```
anniefu, line 17: In its final form, is the engine Search is going to handle converting raw query text => keywords and synonyms?

If so, I'd leave those out of the parameters for any class functions, since the caller of the functions shouldn't need to know that keyword extractoin & synonym lookup is happening behind the scenes.

If needed for prototype, you can hardcode an internal map of { raw query text : keyword lists } and { keywords : synonyms }.

prototyping/testing/searchTest.py
anniefu, line 1: I'd recommend using the [unittest](https://docs.python.org/2/library/unittest.html) framework for this (or select a different python unit test framework).

prototyping/Search.py
Domene99: Hadn't thought about that thanks

prototyping/Search.py
Domene99: No, there will be a utility function that returns the keywords from the query text and another that returns a list of synonyms from the list of keywords. Should be the next pr.

prototyping/testing/searchTest.py
Domene99: @jaimehisao and I chose pytest because of it's ease of use with github actions. We did look into the unittest framework but we didn't see why we should choose that one. Could you please explain the shortcomings that pytest has?

searchEngine/testing/searchTest.py
```diff
@@ -0,0 +1,69 @@
+import os
+import sys
+import inspect
+
+current_dir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
+parent_dir = os.path.dirname(current_dir)
+sys.path.insert(0, parent_dir)
```
ssmall, line 7: I've never seen this done before, seems pretty hacky even for a prototype. Is this common practice for testing in Python?

searchEngine/SearchEngine.py
```diff
@@ -0,0 +1,55 @@
...
+            score = 0
+            article_frequency = article.keywords
+            for keyword in self.keywords:
+                score += article_frequency.get(keyword, 0) * self.keywords_weight
+
+            for synonym in self.synonyms:
+                score += article_frequency.get(synonym, 0) * self.synonyms_weight
+
+            if score != 0:
+                score_per_article[article.id] = score
```
ssmall, line 53: If I understand correctly how this algorithm works, it seems that words which have more synonyms will always have more weight.

For example, if word A has only one available synonym, and word B has five available synonyms, the effective significance of word B will be 5x that of word A.

This is important for proper nouns, which are unlikely to have synonyms but (at least intuitively) are probably far more relevant to the search results.

ssmall, line 53: Have you considered also scoring articles based on how many of the keywords appear in the article? Consider the following example:
```
Keywords: car, seatbelt
Article 1 contains: car x 500, seatbelt x 0
Article 2 contains: car x 50,  seatbelt x 50
```

If I understand the algorithm correctly, Article 1 will be the highest-ranking result even though it never mentions the word "seatbelt".

searchEngine/SearchEngine.py
```diff
@@ -0,0 +1,55 @@
...
+        synonyms_weight: The weight that a synonym has on the scoring algorithm
+            Optional: default argument as 1
+    """
+    def __init__(self, articles):
+        """
+        Inits the search engine by loading all articles
+        """
+        self.articles = articles
+
+    def query(self, keywords, synonyms, keywords_weight=1, synonyms_weight=1):
```
anniefu, line 23: Do you foresee the same query being called multiple times often? If not, you might not need to store the keywords/synonyms here, and you can merge `query` and `score_articles` to be one function. 

That way your class interface involves a few less calls:

```python
    search = SearchEngine(articles)
    assert result == search.query(keywords_article_1, synonyms_article_1)
    assert result == search.query(keywords_article_2, synonyms_article_2)
```

searchEngine/testing/searchTest.py
```diff
@@ -0,0 +1,69 @@
...
+import inspect
+
+current_dir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
+parent_dir = os.path.dirname(current_dir)
+sys.path.insert(0, parent_dir) 
+from SearchEngine import SearchEngine
+from MockArticle import Article
+
+# Assuming articles already come in lemma form, lemma form created manually using cloud natural language
+article_1 = Article(1, 1,"El conductor y ocupante de bicicleta, bicimoto o triciclos poder utilizar de preferencia casco de proteccion para el seguridad.",
```
ssmall, line 12: This example suggests to me that, especially in a language such as Spanish, you will probably need to expand your definition of keywords to include [bigrams and trigrams][]. In this example, `casco de proteccion` seems like an important keyword, but your current approach to the prototype does not represent it.

[bigrams and trigrams]: https://towardsdatascience.com/understanding-word-n-grams-and-n-gram-probability-in-natural-language-processing-9d9eef0fa058

searchEngine/SearchEngine.py
```diff
@@ -0,0 +1,55 @@
...
+    Attributes:
+        keywords: A list of keywords
+        synonyms: A list of synonyms for the keywords
+        articles: A list of articles
+        keywords_weight: The weight that a keyword has on the scorig algorithm
+            Optional: default argument as 1
+        synonyms_weight: The weight that a synonym has on the scoring algorithm
+            Optional: default argument as 1
+    """
+    def __init__(self, articles):
```
anniefu, line 17: Something to consider is if you want `keywords_weight` and `synonyms_weight` to be a configuration of the entire `SearchEngine`, or a per-query configuration.

In general, allowing more fine-grained configuration is nice because it's more flexible, but too much flexibility can also cause end users to become overwhelmed or confused. So you just want to find the balance that you think works best.

anniefu:
oh actually, according to Google python style guide, you'll want to name your files (modules) and directories (packages) using `lower_with_under.py` : https://google.github.io/styleguide/pyguide.html#3163-file-naming

searchEngine/MockArticle.py
```diff
@@ -0,0 +1,16 @@
+class Article:
+    """
+    Has all the attributes that an article table will have in the db
+    Attributes:
+        id: given id in db
+        number: article number extracted from the pdf
+        content: article's full text
+        keywords: frequency map of the keywords from the content
```
jaimehisao, line 8: Nice! I guess we are both going to use this class. I'm still not sure about the folder structure but maybe we should put it one level up?

searchEngine/testing/searchTest.py
jaimehisao: There seems to be a lot of ways to import a parent directory, [this](https://docs.pytest.org/en/stable/goodpractices.html) looks like a good resource for a bit cleaner code.

searchEngine/SearchEngine.py
```diff
@@ -0,0 +1,55 @@
+from MockArticle import Article
```
oscardavidrm, line 1: +1 to Annie's comment of following Google's snake convention on file naming.

searchEngine/MockArticle.py
oscardavidrm: Yeah, probably both of your services could be grouped in a subfolder of the project.

searchEngine/testing/searchTest.py
Domene99: Yeah that was my bad I got really confused between python 2 and 3 due to the __init__.py so I just used the first thing that worked

searchEngine/SearchEngine.py
Domene99: Yeah, I hadn't thought about that. Seems like a real issue. Maybe the synonym weight attribute can change depending on the amount of synonyms so that the effective significance is always the same, but that seems more like a patch than a design decision.

Domene99: I see what you mean, but I think that taking into consideration all of these factors is beyond this pr. It's good to keep in mind which is why I created the possible path expansions section in my design doc, which talks about additional cases to take into consideration to improve the search algorithm. Which I plan on adding all or most of them one by one.

searchEngine/testing/searchTest.py
Domene99: This is something I sort of mentioned in our original design doc. I mentioned context analysis, not n grams, but a close enough concept. I'm not sure if it fits the scope of this project especially since there are so many possible improvements the search algorithm needs, but I will add it to possible improvements.

Domene99: Apparently context analysis uses n grams. https://www.lexalytics.com/lexablog/context-analysis-nlp

searchEngine/MockArticle.py
Domene99: moved it

searchEngine/SearchEngine.py
ssmall: Probably fine to leave as-is for the prototype and see how it performs.

PR #11, opened 2020-07-24
server (setup) Setup basic Apollo Server running on Node.js

oscardavidrm:
+ config/index.js loads necessary env variables to run the server instance.
+ graphql/index.js instantiates an Apollo Server instance with basic configuration
and stitching between ypeDefs and resolvers loading.
+ Added *type Article* to typeDefs and a simple mock query that fetches a list of
articles based of an incoming search. (article/queries.js, resolvers/index.js, article.graphql, index.graphql)
+ server/index.js initializes the server and binds it to a port depending on the environment to listen
for incoming requests.
+ Added several dependencies to build the server instance into package.json.


server/graphql/resolvers/index.js
```diff
@@ -0,0 +1,10 @@
+// Test
```
Domene99, line 1: What does this comment mean. What is it testing specifically?

anniefu:
Doesn't this duplicate some of the changes from #11? Is that intentional?

PR #12, opened 2020-07-24
app (setup) Create initial Web App and add prototype template.

oscardavidrm:
+ Add .gitignore to ignore files in both projects.
+ Basic jsconfig.json to handle relative imports within the application.
+ Added needed React dependencies to package.json to setup the basic template and
added Material UI as the component library.
+ Personalization of app/public/ folder to customize meta tags and displayable icons
on mobile and desktop devices. Added within a /static folder to serve static files
such as the logo.
+ Added routing and code chunking serving logic in App.js
+ Apollo Client setup within apollo/index.js to support queries to the backend service.
+ Added initial components/layout template to contain the entire application components.[
+ Added config/index.js to load env variables needed to setup the Apollo Client.
+ src/index.css contains default config to handle global CSS styles.
+ Added ApolloProvider and BrowserRouter to src/index.js to inject the Apollo Client and
enable routing within the application.
+ src/serviceWorker.js contains a default configuration with every create-react-app app.
+ Added view views/search-engine as the default root page offering a ready-to-query search
engine to make request that are routed to the server.


app/src/views/search-engine/components/articles-list/index.js
```diff
@@ -0,0 +1,97 @@
...
+                </Typography>
+                <Typography
+                  variant="caption"
+                  color="textSecondary"
+                  component="p"
+                  alignLeft
+                  noWrap
+                  className={classes.keywords}
+                >
+                  {`Palabras clave: ${article.keywords.join(', ')}`}
```
Domene99, line 53: Are we displaying the keywords to the user?

app/src/views/search-engine/components/search-bar/index.js
```diff
@@ -0,0 +1,28 @@
...
+      <Toolbar>
+        <div className={classes.search}>
+          <Input
+            type="search"
+            placeholder="Buscar..."
+            classes={{
+              root: classes.inputRoot,
+              input: classes.inputInput,
+            }}
+            onChange={({ target: { value } }) => setSearch(value)}
```
Domene99, line 19: Could we also please add an actual button to trigger the search. It is not inherent for everyone to click the enter key. It helps greatly in terms of accessibility for the elderly and visually challenged people the addition of buttons. I know we are not that focused on this demographic but a simple button is a good investment in terms of time v.s. accessibility created.

app/src/views/search-engine/components/articles-list/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+const ArticlesList = ({ articles }) => {
+  const [currentArticle, setCurrentArticle] = useState(undefined);
+  const classes = styles();
+
+  return (
+    <Grid container spacing={3}>
+      {articles.map((article) => {
+        const contentPreview =
+          article.content.length > 200
```
Domene99, line 26: Please make the article preview length a constant

app/src/views/search-engine/components/search-bar/index.js
```diff
@@ -0,0 +1,28 @@
...
+          />
+        </div>
+        <Avatar alt="Reglamento Vial" src="/static/icon-bg.png" className={classes.avatar} />
```
Domene99, line 22: Maybe this image could be the trigger button as well, it already seems to be in the right position anyways and it currently serves no other purpose.

app/src/views/search-engine/components/articles-list/index.js
```diff
@@ -0,0 +1,97 @@
...
+          <DialogTitle>{`Artículo ${currentArticle.number}`}</DialogTitle>
+          <DialogContent dividers>
+            <DialogContentText style={{ whiteSpace: 'pre-wrap' }}>
+              {currentArticle.content}
+            </DialogContentText>
+          </DialogContent>
+          <DialogActions>
+            <Button
+              variant="contained"
+              autoFocus
```
Domene99, line 84: I think the autoFocus is the bug I found where the first time you open an article the ok btn looks kind of weird

Domene99: In depth issue explanation. The ok button has this sort of breathing animation when a user first opens an article in the session. You have to hover or click the button to get rid of it. After the first time this does not happen until the clients session ends.

app/src/views/search-engine/components/articles-list/index.js
oscardavidrm: Of course this is just me playing around with the design to see what fits best, but if you think about it, showing the user what keywords relate to each article, may educate them in doing more efficient queries. Since it may help them understand what is the algorithm looking for depending on what they search for.

app/src/views/search-engine/components/search-bar/index.js
oscardavidrm: Sure, although, the second you start typing, the search bar is ready to perform a query, so, the second you finish typing, the request is automatically triggered, there's no need for any manual triggering, rendering the suggested button useless.

app/src/views/search-engine/components/articles-list/index.js
oscardavidrm: I see what you mean, I'll go ahead and remove that attribute since it is indeed causing that visual bug. It is most likely that it has to do with Material UI.

app/src/views/search-engine/components/search-bar/index.js
oscardavidrm: Yeah, I'll go ahead and upgrade that image to a trigger. It doesn't hurt anyone.

app/src/apollo/index.js
```diff
@@ -0,0 +1,18 @@
...
+      fetchPolicy: 'network-only',
+      errorPolicy: 'all',
+    },
+    mutate: {
+      errorPolicy: 'all',
+    },
+  },
+});
+
+export default client;
```
jaimehisao, line 18: So we are still going to use GraphQL and Apollo?

app/src/views/search-engine/components/articles-list/index.js
```diff
@@ -0,0 +1,96 @@
...
+                  {`Palabras clave: ${article.keywords.join(', ')}`}
+                </Typography>
+              </CardContent>
+              <CardActions>
+                <Chip
+                  size="small"
+                  color="secondary"
+                  label={`${article.minutesToRead} mins`}
+                  className={classes.minutesToRead}
+                />
```
jaimehisao, line 61: What is `minutesToRead`?

app/src/views/search-engine/components/search-bar/index.js
jaimehisao: This is true for many apps, that the searches are done automatically, or at least suggested automatically, but all of this apps still have some kind of button to press to force the search. Maybe put it inside the text box but translucent?

app/src/apollo/index.js
oscardavidrm: Yes, if no one has any issue with it, I elaborated on my design doc pros and cons of the tech stack.

I did however refactored and integrated the backend service into App Engine, but those changes are available on a separate PR.

app/src/views/search-engine/components/articles-list/index.js
oscardavidrm: The list of attributes are described in our [Data Model and Query analysis](https://docs.google.com/document/d/1OxBZDq-5sbgGk02-GkZE7UdDEd_ouoopLU3hSAEJKH8/edit) document.

But it is the average read time per article.

app/src/views/search-engine/components/search-bar/index.js
oscardavidrm: Yeah, I'll be adding a search icon to force any request.

app/src/views/search-engine/components/search-bar/index.js
oscardavidrm: I'll be adding an additional search icon adjacent to the search bar.

app/src/constants/misc.js
```diff
@@ -0,0 +1,3 @@
+const DEFAULT_ARTICLE_PREVIEW_SIZE = 200;
```
anniefu, line 1: Probably want to name this something more specific than "misc.js". Unless you're planning on just putting all constants in the same file, in which case you might just want a single "constants.js" file.

app/src/views/search-engine/components/articles-list/styles.js
```diff
@@ -0,0 +1,19 @@
+import { makeStyles } from '@material-ui/core/styles';
```
anniefu, line 1: Is it possible with this library to read the CSS from a .css file?

Google style guide recommends putting CSS/styling in its own file: https://google.github.io/styleguide/htmlcssguide.html#Separation_of_Concerns

Seems like React is for separating it into .css files too:
https://reactjs.org/docs/faq-styling.html#what-is-css-in-js

oscardavidrm: Yeah, this way is based of Material UI way of [encapsulating classes](https://material-ui.com/styles/api/#makestyles-styles-options-hook). I was just following Material UI way of using CSS classes.

app/src/App.js
```diff
@@ -0,0 +1,24 @@
...
+const SearchEngine = Loadable({
+  loader: () => import('./views/search-engine'),
+  loading: TopBarProgress,
+});
+
+const App = () => {
+  return (
+    <Switch>
+      <Layout>
+        <Route exact path="/" component={SearchEngine} />
```
Domene99, line 17: Do you need the path="/", I would guess that is the default path and removing it makes no difference

app/src/constants/misc.js
oscardavidrm: Yeah that's true. Probably we will not be having so many constants, so it makes sense to change the file name.

app/src/App.js
oscardavidrm: For each component/view you create, you must specify the path that references that component. Plus this is good cause if the user receives an incorrect path, they will be redirected by default to the root folder.

You can always check react-router-dom library for reference. Here is a [quick guide to get started](https://reactrouter.com/web/guides/quick-start).

app/src/serviceWorker.js
```diff
@@ -0,0 +1,141 @@
...
+// register() is not called by default.
+
+// This lets the app load faster on subsequent visits in production, and gives
+// it offline capabilities. However, it also means that developers (and users)
+// will only see deployed updates on subsequent visits to a page, after all the
+// existing tabs open on the page have been closed, since previously cached
+// resources are updated in the background.
+
+// To learn more about the benefits of this model and instructions on how to
+// opt-in, read https://bit.ly/CRA-PWA
```
ssmall, line 11: It looks like this code was copied from somewhere. Is it auto-generated by create-react-app?

oscardavidrm: Yes, that's correct. [Here's](https://create-react-app.dev/docs/making-a-progressive-web-app/) the reference to the default serviceWorker.js functionality including its benefits and limitations. 

As a summary, it helps improve caching performance to smoothen the user experience. I wouldn't mind leaving it nor removing it. Wdyt?

ssmall: Is it currently used in the code, or do you have plans to use it for this project? If it's unused code, it should probably just be deleted. Otherwise, it's fine to include here.

oscardavidrm: Yes I thought we could use it since it does help smoothen the user experience quite a lot. But it gets "activated" once you deploy to a production environment.

PR #13, opened 2020-07-27
frontend (setup) Add initial app.yaml and cloudbuild.yaml config …

oscardavidrm:
…files.

+ The app.yaml file is consumed by App Engine to setup paths and general configuration.
+ The cloudbuild.yaml is consumed by Cloud Build to understand how to build the webapp before deploying it to App Engine.


app/cloudbuild.yaml
```diff
@@ -0,0 +1,11 @@
+steps:
+  - name: node:10.16.3
+    entrypoint: yarn
+    args: [install, --production]
+  - name: node:10.16.3
+    entrypoint: yarn
+    args: [build]
+    env:
```
anniefu, line 8: Where does this URL come from?

anniefu: Hmm I don't think this was answered?

oscardavidrm: By the URL you mean `node:10.16.3`?
If that's the one you mean, [here](https://hub.docker.com/layers/node/library/node/10.16.3/images/sha256-925162bdc4b23422763cb2d08ecf1661bbbd856ab54b959517924054e17813fc?context=explore) is the link.

I am having a hard time to find the specific reference, but I did look into [Google's official Cloud Builders](http://gcr.io/cloud-builders) docs and found out that the specific `node:yarn` version available was very limited in terms of available tools and was having a hard time performing the steps. So I found this official Node image available on Docker Hub.

Domene99: I think the url would be the react app http

oscardavidrm: Oh okay, I see.

That link points at a personal service I have running on App Engine. I used it to demo to you guys.

But definitely this link should be changed for any link that gets generated when we setup the App Engine service.

app/cloudbuild.yaml
```diff
@@ -0,0 +1,11 @@
...
+      - REACT_APP_HTTP_URI=https://graphql-dot-keen-apogee-250800.uc.r.appspot.com
```
ssmall, line 9: This should probably be passed in as a [substitution][] so that it can be used in multiple environments.

[substitution]: https://cloud.google.com/cloud-build/docs/configuring-builds/substitute-variable-values

oscardavidrm: Oh yeah! That's exactly what I was looking for. This is great!

PR #15, opened 2020-07-28
Set up Super Linter

ssmall:
Using guide at https://github.com/marketplace/actions/super-linter


jaimehisao:
Nice addition! I was about to mention the token issue

PR #17, opened 2020-08-03
Set up NLP search services MVP

Domene99:
For every service added:
- utils functions
- constants file
- main service function
- Docker fail

Closing #16 
Closing #20


constants.py
```diff
@@ -0,0 +1,36 @@
+from mock_article import Article
```
anniefu, line 1: This is just for testing, right? Can it be moved underneath the testing folder?

workflow.py
```diff
@@ -0,0 +1,27 @@
...
+
+    for token in short_gresponse['tokens']:
+        keywords.append(token['lemma'])
+
+    lan = short_gresponse['lan']
+    synonyms = []
+    if lan == 'es':
+        for word in keywords:
+            for synonym in utils.create_synonym_list_esp(word):
+                synonyms.append(synonym)
```
anniefu, line 18: Python has a built-in function for appending lists to lists called `extend`, so you don't have to do this extra for loop for synonyms: https://www.geeksforgeeks.org/append-extend-python/

workflow.py
```diff
@@ -0,0 +1,27 @@
...
+    short_gresponse = utils.from_gcloud_remove_extra(utils.gcloud_syntax_extraction(query))
+
+    keywords = []
+
+    for token in short_gresponse['tokens']:
+        keywords.append(token['lemma'])
+
+    lan = short_gresponse['lan']
+    synonyms = []
+    if lan == 'es':
```
anniefu, line 15: Are you guys planning on supporting languages other than spanish for now? If not, you can probably move this language check to near the top of this function as a base case, and return early if it's not spanish.

workflow.py
```diff
@@ -0,0 +1,27 @@
...
+        for word in keywords:
+            for synonym in utils.create_synonym_list_esp(word):
+                synonyms.append(synonym)
+
+    return search.query(keywords, synonyms)
+
+def main():
+    search = SearchEngine(constants.ARTICLES, keywords_weight=2)
+    print(query_workflow(search, constants.QUERY_ARTICLE_1))
```
anniefu, line 24: This is a bit awkward, because it forces the end-user to figure out their own query_workflow to extract keywords & synonyms, but really the SearchEngine class can handle that all for you.

I would  flip around your `search.query` and `query_workflow` functions, so that your API is simply `search.query("this is a raw string query")`, and the first thing `search.query` does is invoke a helper method like SearchEngline class helper method like `query_workflow` that extracts keywords and synonyms.

utils.py
```diff
@@ -0,0 +1,67 @@
...
+            optional: Default value 5
+    Returns: a list of synonyms
+    """
+    synonyms = []
+    for word in word_arr:
+        payload = requests.get(SPANISH_API_URL+word)
+        word_synonyms = payload.json()['sinonimos']
+        if len(word_synonyms) > max_synonyms:
+            word_synonyms = word_synonyms[:max_synonyms]
+        for synonym in word_synonyms:
```
anniefu, line 64: Also good opportunity to use python's built-in `extend` function here

utils.py
```diff
@@ -0,0 +1,67 @@
...
+    client = language.LanguageServiceClient()
+    # pylint: disable=no-member
+    document = types.Document(
+        content=text,
+        type=enums.Document.Type.PLAIN_TEXT)
+    response = client.analyze_syntax(document=document)
+
+    return response
+
+def from_gcloud_remove_extra(gcloud_response):
```
anniefu, line 25: I would flip the functionality of `from_gcloud_remove_extra` and `gcloud_syntax_extraction` too.

The high-level goal here is really just a function that takes in a query and returns keywords, so your function really probably look something like `extract_keywords(query)`, and behind-the-scenes that function calls `gcloud_syntax_extraction` and does some parsing of the response.

But the fact that you use the gcloud syntax extraction & remove certain parts of the response is an implementation detail, and something the caller of the function doesn't need to know about, so you don't want to expose it in the function stub.

It also makes it easier to switch out your implementation later, because if you choose to use AWS syntax extractin or something like that, anyone calling `extract_keywords(query)` won't break, but anyone using `utils.from_gcloud_remove_extra(utils.gcloud_syntax_extraction(query))` will.

utils.py
```diff
@@ -0,0 +1,67 @@
...
+        if part_of_speech_tag in KEY_PARTS_OF_SPEECH:
+            token_data = {'word' : token.text.content, 'lemma' : token.lemma,
+                          'part_of_speech' : part_of_speech_tag}
+            tokens_shortened.append(token_data)
+
+    response = {'lan' : gcloud_response.language, 'tokens' : tokens_shortened}
+
+    return response
+
+def create_synonym_list_esp(word_arr, max_synonyms=5):
```
oscardavidrm, line 49: You could probably refactor the default `max_synonyms` value into a constant so that if you ever decide to reference it somewhere else, you don't have to worry about matching the default value everywhere in your code.

utils.py
```diff
@@ -0,0 +1,67 @@
...
+    Attributes:
+        word_arr: The word list which we wish to find synonyms for
+        max_synonyms: The maximum number of synonyms per word in word_arr
+            optional: Default value 5
+    Returns: a list of synonyms
+    """
+    synonyms = []
+    for word in word_arr:
+        payload = requests.get(SPANISH_API_URL+word)
+        word_synonyms = payload.json()['sinonimos']
```
oscardavidrm, line 61: How would this handle incoming errors from the `SPANISH_API_URL`? 

I guess given the proper scenario, you'd always expect to get a `sinonimos` attribute in the response payload, but what would happen if an error gets returned? Enclosing this section of code in a try/catch block would be ideal, at least to log the error and avoid the program from breaking.

utils.py
```diff
@@ -0,0 +1,67 @@
...
+    the language, the words found, the lemma version of said words, the part of speech for said word
+    and any word that is not part of the key parts of speech constants.
+    Attributes:
+        gcloud_response: the response given by the natural language api
+    Returns:
+        An object with the language, the keywords as defined by the design doc,
+        the lemma and part of speech for said keywords
+    """
+    tokens_shortened = []
+    for token in gcloud_response.tokens:
```
oscardavidrm, line 37: I haven't looked into the API, but is there any way for you to model the response format on the same query? Like trying to only fetch the information you want rather than downloading all the payload and curating it in here.

Can you pass any sort of params or so to define how the response should look like?

I am asking because you've been diving deep into this documentation, so I'm pretty sure you have better insights.

I think that if you find anything of the sort, you could reduce CPU cycles and overall computation and iterations you have to perform to extract what matters to you from the response. If it's not possible, then I guess this is the way to go, but it'd be good to know if there's anything available.

utils.py
Domene99: Doesn't seem like it will work as is because of the way the data is structured. Could use a lambda expression or list comprehension but it would just be more unreadable.
[
{sinonimo: perro},
{sinonimo: canino}
]

utils.py
Domene99: Hmmmm I think it's a matter of if I'll ever use it which makes me think this is a typical case of "you'll never use it". For now I think this gives me some flexibility and if I ever do need the max synonyms anywhere else I'll have to change it. But I don't know where I could use it

utils.py
Domene99: All examples of this api call I've seen seem to indicate that there's no way to filter down the response

services/keywords/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+# https://hub.docker.com/_/python
+FROM python:3.7-slim
+
+# Copy local code to the container image.
+ENV APP_HOME /app
+WORKDIR $APP_HOME
+COPY . .
+
+# Install production dependencies.
+RUN pip install functions-framework
```
jaimehisao, line 11: This is unnecessary since `functions-framework` is already specified inside the `requirements.txt`
Having an extra `run` command generates an extra docker layer too.

services/search/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+# https://hub.docker.com/_/python
+FROM python:3.7-slim
+
+# Copy local code to the container image.
+ENV APP_HOME /app
+WORKDIR $APP_HOME
+COPY . .
+
+# Install production dependencies.
+RUN pip install functions-framework
```
jaimehisao, line 11: add `function-framework` to `requirements.txt` so you save the extra layer and also have an accurate outline of the required dependencies to run this.

services/search/testing/test_search_engine.py
```diff
@@ -0,0 +1,27 @@
+from search_engine import SearchEngine
```
jaimehisao, line 1: Nice tests! I like that they are cleaner!

services/search/search_engine.py
```diff
@@ -20,7 +24,39 @@ def __init__(self, articles, keywords_weight=1, synonyms_weight=1):
...
+        """
+        Extracts keywords, from keywords endpoint, 
+        and synonyms, from local function,
+        to call search query and calculate the score for articles
+        Args:
+            query: query string
+        Returns:
+            A map of the score of every article
+        """
+        url = "http://localhost:8080"
```
jaimehisao, line 37: Can you use the `KEYWORDS_ENDPOINT` in the constants file here?

services/keywords/Dockerfile
Domene99: Huh yeah, that seems reasonable. I might have missed it when reading it, but this is from the [example dockerfile](https://github.com/GoogleCloudPlatform/functions-framework-python/blob/master/examples/cloud_run_http/Dockerfile), and within the same paragraph it tells you to put it in the requirements.txt https://github.com/GoogleCloudPlatform/functions-framework-python#cloud-runcloud-run-on-gke

services/search/search_engine.py
Domene99: HAHAHHA silly me

services/keywords/Dockerfile
Domene99: I'm going to leave the conversation open if anyone has an idea on why the functions framework left both

jaimehisao: I'm guessing that they did it that way so that there's a layer with that Python dependency and they "save" time when they rebuild the image with a changed `requirements.txt`. But in our case that makes no difference as our Cloud Build will always rebuild the Docker Image with each trigger.

services/keywords/main.py
```diff
@@ -0,0 +1,13 @@
...
+def get_keywords_service(request):
+    """
+    Parses http json request and returns keywords and language of text
+    Args:
+        request: http POST body request as json
+    Returns:
+        json of language detected and keywords for text
+    """
+    json = request.get_json()
+    text = json["text"]
```
anniefu, line 13: I would handle the case that "text" does not exist in incoming JSON.

services/keywords/utils.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    try:
+        client = language.LanguageServiceClient()
+        # pylint: disable=no-member
+        document = types.Document(
+            content=text,
+            type=enums.Document.Type.PLAIN_TEXT)
+        response = client.analyze_syntax(document=document)
+    except exceptions.GoogleAPICallError as err:
+        return {"error": {"type": "GoogleAPICallError",
```
anniefu, line 22: So this `gcloud_syntax_extraction` shouldn't need to know about your API's agreed-upon `error` return. I'd recommend deferring that to the calling func `extract_keywords` and just re-raising the error here (or not catching it at all). That will also save you a lot of copy-pasting.

services/keywords/utils.py
```diff
@@ -0,0 +1,60 @@
...
+        return {"error": {"type": "TypeError", 
+            "message": getattr(err, 'message', str(err))}}
+    except ValueError as err:
+        return {"error": {"type": "ValueError", 
+            "message": getattr(err, 'message', str(err))}}
+
+
+    return response
+
+def extract_keywords(text):
```
anniefu, line 34: [nit] This is pretty much your core service logic vs. just a util func. Maybe move this to `main.py` or rename this file to something more specific, like `extract.py`

services/keywords/main.py
```diff
@@ -0,0 +1,13 @@
+import utils
+
+def get_keywords_service(request):
```
anniefu, line 4: Please add tests, you can fake out an incoming flask request by doing:

```
def make_flask_request(json):
    return flask.Request.from_values(content_type="application/json", data=json)
```

You'll probably want to use a test library like `unittest.mock` to help mock out `gcloud_syntax_extraction`: https://fgimian.github.io/blog/2014/04/10/using-the-python-mock-library-to-fake-regular-functions-during-tests/

Feel free to use a different `pytest` one though.

services/search/constants.py
```diff
@@ -0,0 +1,5 @@
+SPANISH_API_URL = "http://sesat.fdi.ucm.es:8080/servicios/rest/sinonimos/json/"
+
+KEYWORDS_ENDPOINT = 'http://localhost:8080'
```
anniefu, line 3: Instead of having this be a constant, please read this from a local environment variable when your server starts up, so it's possible to switch between local testing and production. You can pass the envvar to your docker container when you run it using the `docker run --env` flag: https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file

(And GKE has a way to pass it to docker too, but you can defer that to whoever is setting up GKE)

services/search/main.py
```diff
@@ -0,0 +1,23 @@
+import testing.test_constants as constants
```
anniefu: remove?

services/search/main.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Parses http json request and returns list of articles and their
+    scores depending on the input query
+    Args:
+        request: http POST body request as json
+    Returns:
+        list of articles with their respective scores
+    """
+    json = request.get_json()
+    query = json["query"]
```
anniefu, line 13: Handle error case of incorrect JSON coming in.

services/search/main.py
```diff
@@ -0,0 +1,23 @@
...
+    search = SearchEngine(keywords_weight=2)
```
anniefu, line 14: The "2" should go in a constant, it can be helpful for testing and if anyone wants to use it somewhere else in the future. Otherwise, you end up with multiple hard-coded "2"s which can be hard to track down if someone wants to change it later on.

services/search/utils.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    synonyms = []
+    for word in word_arr:
+        try:
+            payload = requests.get(constants.SPANISH_API_URL+word)
+            payload.raise_for_status()
+        except requests.exceptions.HTTPError as err:
+            return {"error": {"type": "HTTPError", "status_code": err.response.status_code,
+                "message": getattr(err, 'message', str(err))}}
+        except requests.exceptions.RequestException as e:
```
anniefu: Same as with `utils.py` from `keywords/`, you should re-raise the exception and defer formatting the error message to the calling function.

services/search/search_engine.py
```diff
@@ -0,0 +1,99 @@
...
+            target_dict: dictionary to update
+        """
+        for word in words:
+            if word in frequency:
+                for key, value in frequency[word].items():
+                    score = target_dict.get(key, 0) + value * weight
+                    if score != 0:
+                        target_dict[key] = score
+
+    def search_query(self, keywords, synonyms):
```
anniefu, line 78: I think this can be an underscored class-internal function too?

services/search/search_engine.py
```diff
@@ -0,0 +1,99 @@
...
+        response = requests.post(constants.KEYWORDS_ENDPOINT, json=query_text)
+        response = response.json()
+        
+        if 'error' in response:
+            return response
+
+        lan = response['lan']
+        
+        if lan not in constants.SUPPORTED_LANGUAGES:
+            return {'error': {'type': 'inputError',
```
anniefu: Please re-raise error here and defer error-formatting to calling function. That way this class can also be used by other python libraries, and not just HTTP servers that need to return JSON.

services/search/main.py
```diff
@@ -0,0 +1,23 @@
...
+    score_per_article = search.query(query)
+
```
anniefu: Ideally, you can try/catch this call, and then return the formatted error here (You'll see that comment in other places too)

services/search/search_engine.py
```diff
@@ -0,0 +1,99 @@
...
+        query_text = {'text': query}
+        response = requests.post(constants.KEYWORDS_ENDPOINT, json=query_text)
+        response = response.json()
+
+        if 'error' in response:
+            return response
+
+        lan = response['lan']
+
+        if lan not in constants.SUPPORTED_LANGUAGES:
```
anniefu, line 46: Please re-raise error here and defer error-formatting to calling function. That way this class can also be used by other python libraries, and not just HTTP servers that need to return JSON.

services/search/main.py
```diff
@@ -0,0 +1,22 @@
...
+    scores depending on the input query
+    Args:
+        request: http POST body request as json
+    Returns:
+        list of articles with their respective scores
+    """
+    json = request.get_json()
+    query = json["query"]
+    search = SearchEngine(keywords_weight=2)
+    score_per_article = search.query(query)
```
anniefu, line 15: Ideally, you can try/catch this call, and then return the formatted error here (You'll see that comment in other places too

services/keywords/extract.py
```diff
@@ -0,0 +1,53 @@
...
+        the response given by the api call
+    """
+    try:
+        client = language.LanguageServiceClient()
+        # pylint: disable=no-member
+        document = types.Document(
+            content=text,
+            type=enums.Document.Type.PLAIN_TEXT)
+        response = client.analyze_syntax(document=document)
+    except Exception:
```
anniefu, line 21: You can remove the try/catch block anywhere you just re-raise the exception. Exceptions will actually automatically get re-raised all the way up the call stack until either the first function to try/catch it or the program's entrypoint (main). 

Normally, you'd only want to try/catch/re-raise if you wanted to add more context to the error message or if you tried to handle the error and failed, which you do not need in this case.

services/keywords/main.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    json = request.get_json()
+    if "text" not in json:
+        return {"error": {"message": "ValueError: Expected 'text' field in json body is missing"}}
+
+    text = json["text"]
+    try:
+        response = extract.extract_keywords(text)
+    except Exception as e:
+        return {"error": {"message": getattr(e, 'message', str(e)),
```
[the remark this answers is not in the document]
oscardavidrm: +1

I like that you're handling multiple situations to make sure the **response** is complete.

services/search/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+ENV APP_HOME /app
+ENV KEYWORDS_ENDPOINT="http://localhost:8081"
+WORKDIR $APP_HOME
+COPY . .
+
+# Install production dependencies.
+RUN pip install -r requirements.txt
+
+# Run the web service on container startup.
+CMD exec functions-framework --target=search_service
```
oscardavidrm, line 15: Although this defaults to `PORT=8080`, maybe explicitly declaring it is good practice so that anyone who tries to spin-up the container in a local environment, knows which ports should get map.

oscardavidrm: Another option, would be to leave a reference to `functions-framework` so that anyone can look into the library and understand how does it works ootb.

services/search/search_engine.py
```diff
@@ -0,0 +1,100 @@
...
+        Extracts keywords, from keywords endpoint,
+        and synonyms, from local function,
+        to call search query and calculate the score for articles
+        Args:
+            query: query string
+        Returns:
+            A map of the score of every article
+        """
+        query_text = {'text': query}
+        response = requests.post(os.getenv('KEYWORDS_ENDPOINT'), json=query_text)
```
oscardavidrm, line 40: Maybe having a fallback default value would be useful for local development.
```
import constants

KEYWORDS_ENDPOINT = os.getenv('KEYWORDS_ENDPOINT') || constants.KEYWORDS_ENDPOINT
response = requests.post(KEYWORDS_ENDPOINT, json=query_text)
```

services/search/main.py
```diff
@@ -0,0 +1,32 @@
+import traceback  # pylint: disable=import-error
```
ssmall, line 1: is the `pylint disable` necessary here?

services/search/search_engine.py
```diff
@@ -0,0 +1,98 @@
+import os
+import requests  # pylint: disable=import-error
```
ssmall, line 2: why is the `pylint disable` necessary here?

PR #18, opened 2020-08-05
spec (add spec) Add initial backend service YAML spec.

oscardavidrm:
+ Created spec/ folder to group the specs of the different available components in the application.
Added the `apollo-server-spec.yaml` to describe request/response formats from the backend service.

### DIFF
--- a/spec/apollo-server-spec.yaml
+++ b/spec/apollo-server-spec.yaml
@@ -1,18 +1,18 @@
 swagger: "2.0"
 info:
-  description: "This is the backend component API. Visit the link for an interactive playground to execute requests."
+  description: "This is the backend component API exposing a GraphQL server. The [GraphQL spec](https://spec.graphql.org/June2018/) dives deep into the philosophy of GraphQL. This spec models and showcases the request/response formats expected from the backend service."
   version: "1.0.0"
   title: "Road Rules Search Engine"
 tags:
   - name: "Backend service"
 schemes:
-  - "https"
+  - "http"
 paths:
   /:
     post:
       tags:
-        - "Root"
-      summary: "Base path"
+        - "Backend service"
+      summary: "GraphQL endpoint"
       description: "Base path handles any incoming request and re-routes it to the requested resolver."
       consumes:
         - "application/json"
--- a/spec/apollo-server-spec.yaml
+++ b/spec/apollo-server-spec.yaml
@@ -33,10 +33,6 @@ definitions:
   Request:
     type: "object"
     properties:
-      operationName:
-        type: "string"
-        description: "The name of the GraphQL resolver being requested."
-        example: "articles"
       query:
         type: "string"
         description: "A GraphQL request containing required attributes."
--- a/spec/apollo-server-spec.yaml
+++ b/spec/apollo-server-spec.yaml
@@ -1,0 +1,56 @@
+swagger: "2.0"
+info:
+  description: "This is the backend component API. Visit the link for an interactive playground to execute requests."
+  version: "1.0.0"
+  title: "Road Rules Search Engine"
+tags:
+  - name: "Backend service"
+schemes:
+  - "https"
+paths:
+  /:
+    post:
+      tags:
+        - "Root"
+      summary: "Base path"
+      description: "Base path handles any incoming request and re-routes it to the requested resolver."
+      consumes:
+        - "application/json"
+      produces:
+        - "application/json"
+      parameters:
+        - in: "body"
+          name: "body"
+          description: "Any GraphQL request ( query/mutation/subscription )"
+          schema:
+            $ref: '#/definitions/Request'
+      responses:
+        "200":
+          description: "Default response format"
+          schema:
+            $ref: '#/definitions/Response'
+definitions:
+  Request:
+    type: "object"
+    properties:
+      operationName:
+        type: "string"
+        description: "The name of the GraphQL resolver being requested."
+        example: "articles"
+      query:
+        type: "string"
+        description: "A GraphQL request containing required attributes."
+        example: "query articles(search: String! limit: Int) { id number content }"
+      variables:
+        type: "object"
+        description: "An object containing all required variables by the operation."
+        example: { search: "test", limit: 5 }
+  Response:
+    type: "object"
+    properties:
+      data:
+        type: "object"
+      errors:
+        type: "array"
+        items:
+          type: "object"


spec/apollo-server-spec.yaml
```diff
@@ -0,0 +1,56 @@
+swagger: "2.0"
+info:
+  description: "This is the backend component API. Visit the link for an interactive playground to execute requests."
+  version: "1.0.0"
+  title: "Road Rules Search Engine"
+tags:
+  - name: "Backend service"
+schemes:
+  - "https"
```
ssmall, line 9: you will only get `http` out-of-the-box unless you do the work to set up SSL for your service, so I would change this to `http` for now

spec/apollo-server-spec.yaml
```diff
@@ -0,0 +1,56 @@
...
+paths:
+  /:
+    post:
+      tags:
+        - "Root"
```
ssmall, line 14: This doesn't match any of the tags defined above on lines 6-7

spec/apollo-server-spec.yaml
```diff
@@ -0,0 +1,56 @@
...
+      summary: "Base path"
```
ssmall, line 15: Is there a more descriptive name you could give this? perhaps something like "GraphQL endpoint"?

spec/apollo-server-spec.yaml
```diff
@@ -0,0 +1,56 @@
...
+        example: "query articles(search: String! limit: Int) { id number content }"
+      variables:
+        type: "object"
+        description: "An object containing all required variables by the operation."
+        example: { search: "test", limit: 5 }
+  Response:
+    type: "object"
+    properties:
+      data:
+        type: "object"
```
ssmall, line 52: Could `data` also be an `array`?

oscardavidrm: No, in any case you'd receive something like:

```
data: { articles: [...] }
```

spec/apollo-server-spec.yaml
jaimehisao, line 14: Change to Backend Service because it shows up under root instead of Backend Service

spec/apollo-server-spec.yaml
```diff
@@ -0,0 +1,56 @@
...
+        - in: "body"
+          name: "body"
+          description: "Any GraphQL request ( query/mutation/subscription )"
+          schema:
+            $ref: '#/definitions/Request'
+      responses:
+        "200":
+          description: "Default response format"
+          schema:
+            $ref: '#/definitions/Response'
```
jaimehisao, line 31: Are there any other responses that could happen? Or will it always be a `200` response?

spec/apollo-server-spec.yaml
```diff
@@ -0,0 +1,56 @@
+swagger: "2.0"
+info:
+  description: "This is the backend component API. Visit the link for an interactive playground to execute requests."
```
jaimehisao, line 3: nit: which link?

jaimehisao, line 3: A bit of more detail would be nice, or a high level description of the endpoint. Also mention that you use GraphQL

spec/apollo-server-spec.yaml
oscardavidrm: It always defaults to 200. It's a GraphQL standard. Rather than dealing with bunch of codes (and having to handle them separately), whenever an error happens, the **json** response comes with an `errors` attribute containing all errors that occurred during the request.

spec/apollo-server-spec.yaml
oscardavidrm: Oh yeah, it supposedly refers to the url where the backend service is exposed. But for now I'll avoid that part of the description since it's not relevant now.

spec/apollo-server-spec.yaml
```diff
@@ -0,0 +1,56 @@
...
+        - in: "body"
+          name: "body"
+          description: "Any GraphQL request ( query/mutation/subscription )"
+          schema:
+            $ref: '#/definitions/Request'
+      responses:
+        "200":
+          description: "Default response format"
+          schema:
+            $ref: '#/definitions/Response'
```
jaimehisao, line 31: will a 200 response always contain an empty `errors` array?

oscardavidrm: Nope, only if there are errors available.

jaimehisao: Okay, so by what I understand from the YAML is that when an operation is successful it will always return a 200 and this (even when errors happen)
`{
  "data": {},
  "errors": [
    {}
  ]
}`

You can add a response other than a 200 that signals an error and you can return the error message there.

spec/apollo-server-spec.yaml
```diff
@@ -0,0 +1,56 @@
...
+        description: "The name of the GraphQL resolver being requested."
+        example: "articles"
+      query:
+        type: "string"
+        description: "A GraphQL request containing required attributes."
+        example: "query articles(search: String! limit: Int) { id number content }"
+      variables:
+        type: "object"
+        description: "An object containing all required variables by the operation."
+        example: { search: "test", limit: 5 }
```
ssmall, line 47: Are any of these properties required? Or are they all optional?

spec/apollo-server-spec.yaml
```diff
@@ -0,0 +1,56 @@
...
+          schema:
+            $ref: '#/definitions/Response'
+definitions:
+  Request:
+    type: "object"
+    properties:
+      operationName:
+        type: "string"
+        description: "The name of the GraphQL resolver being requested."
+        example: "articles"
```
ssmall, line 39: Why is the operationName necessary? Isn't it encapsulated in the `query`?

spec/apollo-server-spec.yaml
oscardavidrm: No, the GraphQL spec specifically mentions that you always get a 200 response status.

The Apollo Server we're using always returns this code.

spec/apollo-server-spec.yaml
oscardavidrm: Yes, that is true. I've read that that attribute specifically can be ommited. How would I go about declaring it as optional in here?

spec/apollo-server-spec.yaml
oscardavidrm: The search property is mandatory and the limit is optional.

oscardavidrm: Virtually, that is stated with the `!` character defined on each resolver. Although you are not restricted for including or avoiding any of the attributes, you'll just get the appropriate error notifying of the missing required attributes.

PR #21, opened 2020-08-10
Added Parser/DB Service API Spec

jaimehisao:
Putting up my API Spec up for merging after fixing comments on GIST


src/parser/apiSpec.yaml
```diff
@@ -0,0 +1,126 @@
+swagger: "2.0"
+info:
+  description: "API description for the Parser/Database component of the Road Search Engine."
```
Domene99, line 3: I don't think the description should be self aware that it's a description. I'd just remove the word "description"

src/parser/apiSpec.yaml
```diff
@@ -0,0 +1,126 @@
...
+host: "localhost"
+basePath: "/v1"
+schemes:
+- "https"
+- "http"
+paths:
+  /parse:
+    post:
+      tags:
+      - "parser"
```
Domene99, line 17: This doesn't match any tag. I believe it should be something like:
tags:
-name: "parser"
  description: "your description"

jaimehisao: the `tag` is the name of the section where the `/parse` lives

src/parser/apiSpec.yaml
```diff
@@ -0,0 +1,126 @@
...
+      keywords:
+        type: "array"
+        items:
+          type: "object"
+          $ref: '#/definitions/KeywordFrequencyPair'
+  # For Reference Only
+  KeywordFrequencyPair:
+    type: "object"
+    properties:
+      keyword:
```
Domene99, line 112: Could you quickly update the response type please?

jaimehisao: sure!

PR #22, opened 2020-08-10
Parser & Database Component

jaimehisao:
PR that contains the code to connect to the DB and Parse w/HTTP endpoints, still on development and will constantly get updated.


jaimehisao:
Tests are in the works!

src/parser/parser.py
```diff
@@ -47,18 +112,64 @@ def count_articles(pdf_text):
...
+
+        # Check if the file is different, this has to be done in a different open file instance
+        # if not, Python returns some weird errors
+        hasher = hashlib.sha256()
+        with open("regs.pdf", "rb") as pdf_file:
+            file_buffer = pdf_file.read()
+            hasher.update(file_buffer)
+            if hasher.hexdigest() == document_to_parse["hash"]:
+                print("Downloaded document is the same as the one currently stored")
+                return "Succesful Operation - File did not change", 200
```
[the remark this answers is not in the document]
jaimehisao: Right, it is a success code [Mozilla has this page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200)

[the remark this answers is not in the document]
jaimehisao: Also I wanted to specify there that the file did not change, but it shouldn't make a difference vs a normal Successful Operation

src/parser/parser.py
```diff
@@ -47,18 +112,64 @@ def count_articles(pdf_text):
...
+                articleDict = {}
+                articleDict["articleNumber"] = article.number
+                articleDict["text"] = article.text
+                articleDict["wordCount"] = len(article.text.split())
+                articles_in_memory.append(articleDict)
+                print("Adding " + str(articleDict))
+                # Get keywords that relate to this article (Javier's service)
+                """
+                    keywords_service_response = requests.get(
+                    "localhost:8000", params={"text": article.text}
```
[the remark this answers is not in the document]
jaimehisao: Yes, I commented this out just to specify that it is going to be used just not now, will change to reflect the `post` change

src/parser/parser.py
```diff
@@ -47,18 +112,64 @@ def count_articles(pdf_text):
...
+                    ).json()
+                """
+                keywords_service_response = keywordMock.get_keywords(article.text)
+                keywords_service_response = json.loads(keywords_service_response)
+                split_article = article.text.split()
+                for keyword in keywords_service_response["keywords"]:
+                    frequency = split_article.count(keyword)
+                    if keyword not in keywords:
+                        keywords[keyword] = []
+                    keywords[keyword].append({'articleNumber': article.number, 'frequency': frequency})
```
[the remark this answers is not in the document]
jaimehisao: Will change accordingly

src/parser/parser.py
```diff
@@ -18,11 +53,40 @@ def __init__(self, number, text):
         self.text = text
 
 
+class Keyword:
```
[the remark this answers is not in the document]
jaimehisao: deleted

src/parser/app.py
```diff
@@ -0,0 +1,29 @@
...
+    # return parser.parse_without_database()
+    return parser.parse()
+
+
+@app.route('/articles/byKeywords', methods=['POST'])
+def get_keywords():
+    data = request.get_json()
+    return jsonify(parser.get_articles_that_match_keywords(data['keywords']))
+    # else:
+    #     return "Invalid Request Body", 500
```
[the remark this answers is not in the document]
jaimehisao: my bad, got commented due to debugging, adding that and the if statement too

jaimehisao:
I couldn't comment on your comment (lol) but the prints will become `logger`

src/parser/parser.py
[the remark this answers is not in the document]
Domene99: I meant to ask if returning ", 200" was a common practice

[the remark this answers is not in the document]
jaimehisao: As far as I know, yes. Can @ssmall and @anniefu chime in on this?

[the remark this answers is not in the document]
ssmall: To answer your question: it is more common to keep the business logic (parsing the document) separate from the application logic (determining what HTTP code to return). In this example, that would mean that the return code would be determined in `app.py` instead of here.

This way of keeping the different kinds of logic separate usually makes unit testing a lot easier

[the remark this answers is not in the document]
jaimehisao: I'll separate the logic for this, now, regarding Javier´s question, returning the 200 is fine?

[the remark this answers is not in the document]
ssmall: > regarding Javier´s question, returning the 200 is fine?

If I understand correctly what you're asking, you're asking if `200` is an appropriate HTTP status code to return in this case. Yes, I believe it is.

(That's not what I understood to be Javier's question, however)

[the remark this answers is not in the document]
Domene99: Yeah my question was definitely about determining the return code within app.py not about the definition of status 200

jaimehisao:
Based on Oscar's suggestion (during a GVC) for image building and deployment on GKE, the folder structure was modified.

src/parser/parser.py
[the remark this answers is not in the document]
jaimehisao: I didn't understand at first then

services/parser-database/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY requirements.txt .
+
+# Install dependencies using PIP
+RUN pip install -r requirements.txt
+
+# Copy Files to container
+COPY . .
+
+# Run the Flask App
+CMD [ "python", "./app.py" ]
```
oscardavidrm, line 20: I think you can avoid the `./` notation and just do `app.py` since it's implicit that you're working on the root folder.

services/parser-database/parser.py
```diff
@@ -0,0 +1,230 @@
...
+
+    Returns:
+        [JSON]: extracted keywords
+    """
+    """
+    return requests.post(
+        "localhost:8000", params={"text": text}
+    ).json()
+    """
+    keywords_service_response = keywordMock.get_keywords(text)
```
oscardavidrm, line 178: nit - Snake case for `keywordMock` -> `keyword_mock`

services/parser-database/keywordMock.py
```diff
@@ -0,0 +1,13 @@
...
+
+def get_keywords(text):
+    splited_text = text.split()
+    keywords = {
+        "keywords": [
+            splited_text[random.randint(0, len(splited_text) - 1)],
+            splited_text[random.randint(0, len(splited_text) - 1)],
+        ]
+    }
+    return json.dumps(keywords)
```
oscardavidrm, line 13: Maybe a brief comment to describe what this line does. I am not really sure what `json.dumps` actually does.

services/parser-database/parser.py
```diff
@@ -0,0 +1,230 @@
...
+    def __init__(self, number, text):
+        self.number = number
+        self.text = text
+
+
+articles_in_memory = []
+keywords_in_memory = {}
+
+
+def get_article_by_number(art_num):
```
oscardavidrm, line 62: Wdyt about using a different in-memory DS structure such as a map to store the list of articles in the following manner (I'm doing the example in `Cython++` which is a new programming language I'm working on lol):
```
unordered_map<Int, Article> articles_in_memory;

def get_article_by_number(art_num):
 if (articles_in_memory[art_num]) return articles_in_memory[art_num];
 return "No article matches such ID", 402;
```
But the equivalent of that in Python.
According to this [doc](https://wiki.python.org/moin/TimeComplexity), a `dict` DS in Python has an O(1) access time for retrieving items.

jaimehisao: I don´t think it is worth it to spend time on this when the DB will persist everything at a later stage...

services/parser-database/parser.py
```diff
@@ -0,0 +1,230 @@
...
+logging.basicConfig(level=logging.INFO)
+
+
+class Document:
+    """Class for Regulations Documents"""
+
+    def __init__(self, jurisdiction, shahash, url, last_updated):
+        self.id = None
+        self.jurisdiction = jurisdiction
+        self.shahash = shahash
```
Domene99, line 28: Nice almost palindrome

services/parser-database/parser.py
```diff
@@ -0,0 +1,230 @@
...
+            article_text += " " + pdf_text[i]
+        i += 1
+    return articles
+
+
+mty_document = {
+    "hash": "afafbfbdce8c40924edae00f6ce54f0c639ce42a2" +
+    "c0fbbfa6ab82ea6925827c51",
+    "jurisdiction": "Monterrey",
+    "url": "http://www.guadalupe.gob.mx/wp-content/upl" +
```
Domene99, line 121: I'd leave it as a one-liner. https://google.github.io/styleguide/pyguide.html#32-line-length

services/parser-database/requirements.txt
```diff
@@ -0,0 +1,4 @@
+Flask
+firebase-admin
+utils
+https://github.com/timClicks/slate/archive/master.zip
```
Domene99, line 4: Does this actually work?

Domene99: apparently it doesn't https://stackoverflow.com/questions/16584552/how-to-state-in-requirements-txt-a-direct-github-source

services/parser-database/parser.py
```diff
@@ -0,0 +1,230 @@
...
+        return True
+
+
+def download_file(url):
+    """Downloads a PDF file from the given URL
+
+    Args:
+        url (string): URL to download file from
+    """
+    retriever.get_document(url)
```
Domene99, line 160: I'd move retriever.get_doc into parse(), seeing as how this function isn't adding any value

services/parser-database/parser.py
jaimehisao: HAHAHA I hadn't noticed that

services/parser-database/parser.py
jaimehisao: `pylint` gave me a very hard time with that, but didn't know about the error disable, will add `# pylint: disable=invalid-name`

src/parser/app.py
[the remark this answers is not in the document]
Domene99: Wait did you remove this file? I still had some questions there

services/parser-database/requirements.txt
jaimehisao: Yes it works, lol

src/parser/app.py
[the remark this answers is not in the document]
jaimehisao: No, the directory was changed because of something Oscar and I talked about yesterday

jaimehisao:
> LGTM just please look into github repos within a requirements.txt

This works, I did testing on this and `pip` installs slate with no issues

services/parser-database/parser.py
```diff
@@ -0,0 +1,218 @@
...
+        pdf_text (string): contains the PDF text where articles
+        are to be identified.
+
+    Returns:
+        list: article objects
+    """
+    articles = []
+    article_text = ""
+    article_count = 1
+    i = 0
```
anniefu, line 92: I would just use a for loop?

services/parser-database/parser.py
```diff
@@ -0,0 +1,218 @@
...
+    while i < len(pdf_text) - 1:
+        if (pdf_text[i] == "artículo" or pdf_text[i] == "articulo") and (
```
anniefu, line 94: So, this would be a really opportune time to use re.split (regular expression split), https://stackoverflow.com/questions/2136556/in-python-how-do-i-split-a-string-and-keep-the-separators for two reasons.
1. It can replace this entire function
2. You don't have to re-build the entire article, which allows you to store it in the database as close to it's original form as possible.

I am happy to help you figure out how to write a regexp equivalent of this. They're pretty confusing the first time you use them.

Plus if you have a test for this function first, then we can be extra sure the regex works :)

services/parser-database/parser.py
```diff
@@ -0,0 +1,218 @@
...
+    """Sees if the file is different.
+
+    Args:
+        past_hash (string): hash to compare from
+
+    Returns:
+        [boolean]: [if the file has changed or not]
+    """
+    hasher = hashlib.sha256()
+    with open("regs.pdf", "rb") as pdf_file:
```
anniefu, line 141: This probably shouldn't be hardcoded anymore. I'd expect the function to take the filename in as a parameter so this can be more general purpose.

services/parser-database/parser.py
```diff
@@ -0,0 +1,218 @@
...
+
+def save_keywords_in_memory(keywords, article):
+    """Saves the keywords from an article in memory
+
+    Args:
+        keywords (JSON): contains keywords
+        article (Article): article object
+    """
+    split_article = article.text.split()
+    for keyword in keywords["keywords"]:
```
anniefu, line 191: [nit] lol, this is a confusing one to read. I would say in your `get_keywords` func you should handle the parsing of the response message from Javier's service so that everything that calls `get_keywords` just gets back a list of the keywords.

services/parser-database/parser.py
```diff
@@ -0,0 +1,218 @@
...
+        frequency = split_article.count(keyword)
```
anniefu, line 192: In hindsight, we should have made Javier's service return the keyword counts too since we know he parsed every word... oh well, maybe in the future.

services/parser-database/parser.py
```diff
@@ -0,0 +1,218 @@
...
+        if keyword not in keywords:
```
anniefu, line 193: I assume you meant `if keyword not in split_article` since you're looping through all the keywords in `keywords`?

Either way, I would recommend not adding blank entries in your dictionary, cause that could take up unnecessary memory for words you may never use again.

services/parser-database/parser.py
```diff
@@ -0,0 +1,218 @@
...
+
+# When the time comes to implement the DB, the commented stuff will allow
+# that (with a few additions)
+def parse(document_to_parse):
+    """Parses all PDF documents that are in the DB"""
+    retriever.get_document(document_to_parse["url"])
+    print('File downloaded')
+    if has_file_changed(document_to_parse["hash"]):
+        print('File has changed')
+        with open("regs.pdf", "rb") as pdf_file:
```
anniefu, line 209: should be `document_to_parse`

services/parser-database/parser.py
```diff
@@ -0,0 +1,218 @@
...
+            doc = slate.PDF(pdf_file)
+            final_text = ""
+            for page in doc:
+                final_text += page
+            final_text = final_text.strip().lower().split()
```
anniefu, line 214: You probably don't want to lower() here, because capitalization is an important part of grammar and the text will eventually get shown to end users.

services/parser-database/parser.py
Domene99: If you want to, I had written this one the other day:
_ART[ÍI]CULO.?\d*\.?-?_
You should probably not take my word for it as I only tested it for a few minutes but I think it works. If you need an explanation on what it does ping me.

services/parser-database/parser.py
jaimehisao: where?

services/parser-database/parser.py
jaimehisao: no, if the keyword that is being added does not exist in the list of keywords, it will create an empty list that will be populated in line 195, if the keyword exists, it will not create an empty list. 

The `split_article` allows the program to search for the frequency of the word in the text.

services/parser-database/parser.py
jaimehisao: Yeah, that would've saved some double processing on my end.

services/parser-database/parser.py
jaimehisao: This generated issues when modifying the `i` value.

services/parser-database/parser.py
```diff
@@ -0,0 +1,218 @@
...
+        self.number = number
+        self.text = text
+
+
+articles_in_memory = []
+keywords_in_memory = {}
+
+
+def get_article_by_number(art_num):
+    for item in articles_in_memory:
```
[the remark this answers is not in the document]
jaimehisao: Firestore would've returned me just the article I requested. This was done due to the in-memory implementation that needed to be done ASAP.

services/parser-database/parser.py
jaimehisao: Yes, I know about the regex. I was implementing it but then had to redesign other things and time got in the way.

I also discussed this with Spencer and `re.match()` is probably a better fit. I would definitely need your help on this @anniefu .

services/parser-database/parser.py
jaimehisao: will see how I can adapt my algorithm, with the regex the lower() wouldn't be necessary so definitely something to look into

services/parser-database/parser.py
jaimehisao: good point, went ahead and took the Jurisdiction attribute of the document as the file name + `.pdf`

services/parser-database/parser.py
jaimehisao: true that, makes for a much more cleaner piece of code

services/parser-database/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Set base image, the slim one allows us to have a small build size
+FROM python:3.8-slim
+
+# set the working directory in the container
+WORKDIR /code
+
+# Exposes the port required for operation on GKE
+EXPOSE 8082
```
ssmall, line 8: This port should be configurable via an environment variable named `PORT`

services/parser-database/app.py
```diff
@@ -0,0 +1,42 @@
+from flask import Flask  # pylint: disable=import-error
+from flask import request  # pylint: disable=import-error
+from flask import jsonify  # pylint: disable=import-error
+
+import parser  # pylint: disable=import-error
```
ssmall, line 5: nit: Can the import error be fixed by just creating an `__init__.py`?

services/parser-database/app.py
```diff
@@ -0,0 +1,42 @@
...
+    except Exception:
+        return "Internal Server Error", 500
+    return "Sucessful Operation", 200
+
+
+@app.route('/articles/byKeywords', methods=['POST'])
+def get_keywords():
+    json_request = request.get_json()
+    if "keywords" not in json_request:
+        return {"error": "keywords request body is missing"}, 500
```
ssmall, line 25: This should return `400` since it is an invalid input

services/parser-database/app.py
```diff
@@ -0,0 +1,42 @@
...
+
+@app.route('/articles/<id>', methods=['GET'])
+def get_article_by_number_in_memory(id):
+    """Returns the articlle that matches the ID value
+    accoring to the apiSpec.yaml file"""
+    article = parser.get_article_by_number(id)
+    if article is not None:
+        return jsonify(article)
+    else:
+        return "No article matches such ID", 402
```
ssmall, line 38: I think you mean `404`? Apparently `HTTP 402` means [Payment Required](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/402) ... I didn't realize your microservice required microtransactions 😝

services/parser-database/app.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+if __name__ == '__main__':
+    app.run(debug=False, host='0.0.0.0', port=8082)
```
ssmall, line 42: Per my comment on `Dockerfile`, `port` should be configurable based on the environment variable `PORT`

ssmall, line 42: nit: Consider setting `debug=True` for now. It will likely come in very helpful for the MVP phase.

services/parser-database/parser.py
```diff
@@ -0,0 +1,218 @@
...
+
+# import datetime
+# import requests
+import json
+
+import slate  # pylint: disable=import-error
+
+import retriever  # pylint: disable=import-error
+# import database  # pylint: disable=import-error
+import keywordmock  # pylint: disable=import-error
```
ssmall, line 12: nit: Can these import errors be fixed by adding `__init__.py`?

services/parser-database/parser.py
```diff
@@ -0,0 +1,218 @@
...
+        self.shahash = shahash
+        self.url = url
+        self.last_updated = last_updated
+
+    @staticmethod
+    def from_dict():
+        pass
+
+    def to_dict(self):
+        pass
```
ssmall, line 31: nit: Are these methods necessary?

services/parser-database/parser.py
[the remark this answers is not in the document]
ssmall: Could you just use a `dict` instead?

services/parser-database/parser.py
```diff
@@ -0,0 +1,218 @@
...
+    with open("regs.pdf", "rb") as pdf_file:
+        file_buffer = pdf_file.read()
+        hasher.update(file_buffer)
+        if hasher.hexdigest() == past_hash:
+            return False
+        return True
+
+
+# TODO Change to point to Javier's service
+def get_keywords(text):
```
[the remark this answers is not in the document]
ssmall: +1 to this suggestion

services/parser-database/requirements.txt
```diff
@@ -0,0 +1,4 @@
+Flask
+firebase-admin
```
ssmall, line 2: nit: firebase-admin should probably not be included in this PR

services/parser-database/parser.py
```diff
@@ -0,0 +1,218 @@
...
+        with open("regs.pdf", "rb") as pdf_file:
+            doc = slate.PDF(pdf_file)
+            final_text = ""
+            for page in doc:
+                final_text += page
+            final_text = final_text.strip().lower().split()
+            articles = count_articles(final_text)
+
+            for article in articles:
+                article_to_dictionary(article)
```
ssmall, line 218: Please add unit tests for the methods in this file before merging.

jaimehisao:
PS: Once the PR get's approved I will move methods around in `parser.py` to increase readability by following order of execution. Waiting on that to not mess up any comments.

services/parser-database/app.py
jaimehisao: HAHAHA

services/parser-database/app.py
```diff
@@ -0,0 +1,46 @@
...
+    accoring to the apiSpec.yaml file"""
+    article = get_article_by_number(id)
+    if article is not None:
+        return jsonify(article)
+    else:
+        return {"error": "Article not found with submitted ID"}, 404
+
+
+if __name__ == '__main__':
+    app.run(debug=True, host='0.0.0.0', port=os.getenv("PORT"))
```
ssmall, line 46: You probably need to put a default value for `PORT` somewhere, either in the Dockerfile or here in the code.

services/parser-database/apiSpec.yaml
```diff
@@ -0,0 +1,130 @@
...
+      parameters:
+      - in: "body"
+        name: "body"
+        description: "Keywords that need to be matched with an article in the database"
+        required: true
+        schema:
+          $ref: '#/definitions/ArticleKeywordsQuery'
+      responses:
+        "400":
+          description: "No articles matched keyword query"
```
ssmall, line 46: nit: Does it make sense to return an error code instead of just an empty list? Usually, error codes indicate that something was *wrong* with the request.

services/parser-database/apiSpec.yaml
```diff
@@ -0,0 +1,130 @@
...
+        required: true
+        description: "ID of the requested article"
+        type: "string"
+      responses:
+        "200":
+          description: "successful operation"
+          schema:
+            $ref: "#/definitions/Article"
+        "400":
+          description: "Invalid Article ID"
```
ssmall, line 72: I believe this should be `404`

services/parser-database/apiSpec.yaml
```diff
@@ -0,0 +1,130 @@
...
+        type: "integer"
+        format: "int 64"
+      text:
+        type: "string"
+      timeToRead:
+        type: "integer"
+        format: "int 64"
+      pageNumber:
+        type: "integer"
+        format: "int 64"
```
ssmall, line 89: By default, the properties are considered optional if they are not marked as `required`. IIUC, all these properties must be present, so they should be marked `required`

services/parser-database/apiSpec.yaml
```diff
@@ -0,0 +1,130 @@
...
+      keyword1:
+        type: "array"
+        items:
+          type: "object"
+          $ref: '#/definitions/KeywordFrequencyPair'
+      keywordN:
+        type: "array"
+        items:
+          type: "object"
+          $ref: '#/definitions/KeywordFrequencyPair'
```
ssmall, line 109: What is `keyword1` and `keywordN`?

services/parser-database/apiSpec.yaml
```diff
@@ -0,0 +1,130 @@
...
+      pageNumber:
+        type: "integer"
+        format: "int 64"
+  ArticleKeywordsQuery:
+     type: "object"
+     properties:
+       keywords:
+        type: "array"
+        items:
+          type: "string"
```
ssmall, line 96: Are both of these properties `required`?

services/parser-database/apiSpec.yaml
```diff
@@ -0,0 +1,130 @@
...
+  # For Reference Only
+  KeywordFrequencyPair:
+    type: "object"
+    properties:
+      articleId:
+        type: "string"
+      keyword:
+        type: "string"
+      count:
+        type: "integer"
```
ssmall, line 119: Are all of these properties `required`?

services/parser-database/apiSpec.yaml
```diff
@@ -0,0 +1,130 @@
...
+    type: "object"
+    properties:
+      id:
+        type: "string"
+      city:
+        type: "string"
+      dateOfDownload:
+        type: "string"
+      hash:
+        type: "string"
```
ssmall, line 130: Are all of these properties `required`?

services/parser-database/app.py
```diff
@@ -0,0 +1,46 @@
...
+
+@app.route('/parse', endpoint='parser', methods=['POST'])
+def trigger_parsing():
+    # return parser.parse_without_database()
+    try:
+        parse_all_documents()
+    except Exception as e:
+        print(e)
+        return "Internal Server Error", 500
+    return "Sucessful Operation", 200
```
ssmall, line 22: nit: What happens when this gets called multiple times concurrently? The API spec says it should return `409` but right now it looks like it will just try to parse in parallel for every request.

(I don't think this needs to be addressed in this PR, but it should probably be marked as a TODO and linked to an issue for Post-MVP)

services/parser-database/parser.py
```diff
@@ -0,0 +1,225 @@
...
+def get_articles_that_match_keywords(keywords_list):
+    """Returns articles that match the the given keyword(s)
+
+    Args:
+        keywords_list (list): Keyword(s) to look for
+
+    Returns:
+        list: articles that match such keyword(s)
+    """
+    mnatching_articles = []
```
ssmall, line 63: nit: `mnatching`?

services/parser-database/parser.py
```diff
@@ -0,0 +1,225 @@
...
+    for keyword in keywords_list:
+        if keyword in keywords_in_memory:
+            print(keyword)
```
ssmall, line 66: Should this use `logging` instead of `print`?

services/parser-database/parser.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def parse_all_documents():
+    """Parses all documents that are specified on the DB
+    (For obvious reasons, this won't work as is while the DB is not in use)
+    (But will still parse the hardcoded document)"""
+    for document in document_list:
+        file_name = document["jurisdiction"] + ".pdf"
+        download_file(document["url"], file_name)
+        print('File downloaded')
```
ssmall, line 122: Should this use `logging` instead of `print`?

services/parser-database/parser.py
```diff
@@ -0,0 +1,225 @@
...
+        file_name_to_use (string): [file name for file to save]
+    """
+    retriever.get_document(url, filename_to_use)
+
+
+def parse(document_to_parse, file_name):
+    """Parses all PDF documents that are in the DB"""
+
+    if has_file_changed(document_to_parse["hash"], file_name):
+        print('File has changed')
```
ssmall, line 215: Should this use `logging` instead of `print`?

services/parser-database/tests/test_parse.py
```diff
@@ -0,0 +1,34 @@
...
+    + " articulo 3. velocidad maxima de 10kmh"
+    + " artículo 4.- carros de mas de 3/4 toneladas requieren licencia de chofer"
+    + " artículo 5.- no te pueden infraccionar en navidad"
+    + " artículo 6.- marcar al seguro y al 911 al chocar"
+)
+
+
+def test_article_extraction_by_number_of_articles_extracted():
+    recieved_articles = identify_articles(articles.strip().split())
+    assert len(recieved_articles) == 6
```
ssmall, line 28: I think this test needs to assert on the content of `received_articles`, not just that it's non-empty. The behavior should be deterministic, so you should be able to write a very detailed unit test for this.

services/parser-database/tests/test_parse.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def test_article_extraction_by_article_content():
+    recieved_articles = identify_articles(articles.strip().split())
+    assert recieved_articles[1].text == " los ciclistas se deben de orillar"
+    assert recieved_articles[4].text == " carros de mas de 3/4 toneladas requieren licencia de chofer"
```
ssmall, line 34: This test can be combined with the above test to make one thorough unit test -- it should assert _exactly_ how many articles are parsed and the exact content of each of the parsed articles.

services/parser-database/parser.py
```diff
@@ -0,0 +1,225 @@
...
+    (For obvious reasons, this won't work as is while the DB is not in use)
+    (But will still parse the hardcoded document)"""
+    for document in document_list:
+        file_name = document["jurisdiction"] + ".pdf"
+        download_file(document["url"], file_name)
+        print('File downloaded')
+        parse(document, file_name)
+
+
+def has_file_changed(past_hash, file_name):
```
ssmall, line 126: Missing a unit test for this method.

services/parser-database/parser.py
```diff
@@ -0,0 +1,225 @@
...
+keywords_in_memory = {}
+
+
+def get_article_by_number(art_num):
+    if art_num in articles_in_memory:
+        return articles_in_memory[art_num]
+    return None
+
+
+def get_articles_that_match_keywords(keywords_list):
```
ssmall, line 54: Missing a unit test for this

services/parser-database/parser.py
```diff
@@ -0,0 +1,225 @@
...
+    "Monterrey",
+    "url":
+    "http://www.guadalupe.gob.mx/wp-content/up" +
+    "loads/2019/09/Nuevo-Reglamento-Homologado-1.pdf",
+}
+document_list = []
+document_list.append(mty_document)
+
+
+def parse_all_documents():
```
ssmall, line 115: Missing a unit test for this

services/parser-database/parser.py
```diff
@@ -0,0 +1,225 @@
...
+        "articleNumber": article.number,
+        "text": article.text,
+        "wordCount": len(article.text.split())
+    }
+    articles_in_memory[str(article.number)] = article_dict
+    keywords = get_keywords(article.text)
+    save_keywords_in_memory(keywords, article)
+
+
+def save_keywords_in_memory(keywords, article):
```
ssmall, line 183: Per https://github.com/googleinterns/major-tom/pull/22/files#r470052450 this method probably belongs in a different file

services/parser-database/parser.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def download_file(url, filename_to_use):
+    """Downloads file from the given URL
+
+    Args:
+        url (string): [URL that contains the file to download]
+        file_name_to_use (string): [file name for file to save]
+    """
+    retriever.get_document(url, filename_to_use)
```
ssmall, line 208: What's the value of this one-line method? Can it just be inlined?

services/parser-database/parser.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    extracted_keywords = []
+    nlp_output = keywordmock.get_keywords(text)
+    json_output = json.loads(nlp_output)
+    for keyword in json_output["keywords"]:
+        extracted_keywords.append(keyword)
+    return extracted_keywords
+
+
+def article_to_dictionary(article):
```
ssmall, line 167: Per https://github.com/googleinterns/major-tom/pull/22/files#r470052450 this method probably belongs in a different file

services/parser-database/parser.py
```diff
@@ -0,0 +1,225 @@
...
+    """Downloads file from the given URL
+
+    Args:
+        url (string): [URL that contains the file to download]
+        file_name_to_use (string): [file name for file to save]
+    """
+    retriever.get_document(url, filename_to_use)
+
+
+def parse(document_to_parse, file_name):
```
ssmall, line 211: Missing unit test for this

services/parser-database/parser.py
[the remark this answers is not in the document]
jaimehisao: done

services/parser-database/app.py
jaimehisao: Yes, if I recall a conversation we had, that was post-MVP. I'll go ahead and create a TODO for that.

services/parser-database/parser.py
jaimehisao: I originally took it out so that the parser test could run with a different file. If it was hard coded the test would have to be with the mty file that may change/dissapear

services/parser-database/parser.py
jaimehisao: haha matching

services/parser-database/parser.py
jaimehisao: I do not know how to test this with mocking since I cannot guarantee I will always get the same keywords when mocking Javier's service.

services/parser-database/apiSpec.yaml
```diff
@@ -0,0 +1,134 @@
...
+  ArticleKeywordFrequencyPair:
+    type: "object"
+    properties:
+      articleId:
+        type: "string"
+      frequency:
+        type: "integer"
+    required:
+    - articleId
+    - frequency
```
ssmall, line 118: This response object needs to include the keywords as well, right? Otherwise there is no way to figure out which keyword from the input is being referenced since the input is a list of keywords.

services/parser-database/app.py
```diff
@@ -0,0 +1,47 @@
+import os
+
+from flask import Flask  # pylint: disable=import-error
+from flask import request  # pylint: disable=import-error
+from flask import jsonify  # pylint: disable=import-error
+
+from parser import parse_all_documents  # pylint: disable=import-error
+from connector import get_articles_that_match_keywords  # pylint: disable=import-error
+from connector import get_article_by_number  # pylint: disable=import-error
```
ssmall, line 9: There shouldn't be import errors here ... these files exist in the same directory. Can you please try to fix the import errors rather than ignoring them?

services/parser-database/app.py
```diff
@@ -0,0 +1,47 @@
...
+
+app = Flask(__name__)
+app.config['JSON_AS_ASCII'] = False
+
+
+@app.route('/parse', endpoint='parser', methods=['POST'])
+def trigger_parsing():
+    # return parser.parse_without_database()
```
ssmall, line 17: nit: Can this line be removed?

services/parser-database/connector.py
```diff
@@ -0,0 +1,89 @@
...
+    Args:
+        text (sting): text to extract keywords from
+
+    Returns:
+        [list]: list of extracted keywords
+    """
+
+    extracted_keywords = []
+    request = {'text': text}
+    nlp_output = requests.post(os.getenv("KEYWORDS_SERVICE"), json=request)
```
ssmall, line 41: Per discussion yesterday, I thought this change would be left out for a different PR.

services/parser-database/tests/test_endpoints.py
```diff
@@ -0,0 +1,63 @@
...
+        'wordCount': 165
+    }
+    assert data["articleNumber"] == must_match["articleNumber"]
+    assert data["text"] == must_match["text"]
+    assert data["wordCount"] == must_match["wordCount"]
```
ssmall, line 63: It looks like this is testing at a very high level (by calling the HTTP endpoints), but that actually makes the tests more of an integration test, especially when you get to the point of integrating the Firebase db. It would be better to test each method individually (true _unit_ testing) and not worry about integration testing for now.

jaimehisao: This is being tested individually in the other file, this test was more oriented to testing the endpoints themselves, which was requested to me.

jaimehisao: I was requested to test the HTTP endpoints, but there is no way to test them without calling them in the regular order of execution.

services/parser-database/parser.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    article_dict = {
+        "articleNumber": article.number,
+        "text": article.text,
+        "wordCount": len(article.text.split())
+    }
+    connector.articles_in_memory[str(article.number)] = article_dict
+    keywords = connector.get_keywords(article.text)
+    logging.info('Article ' + str(article.number) + ' assigned keywords')
+    connector.save_keywords_in_memory(keywords, article)
```
ssmall, line 119: There is actually a lot going on in this method that may make it hard to unit test. I would suggest splitting it up into a couple pieces and testing each of those pieces independently. Suggestions for each piece are in separate comments.

services/parser-database/parser.py
```diff
@@ -0,0 +1,136 @@
...
+    """Converts an parsed article to a dictionary and saves it
+
+    Args:
+        article (Article): [article object]
+    """
+    article_dict = {
+        "articleNumber": article.number,
+        "text": article.text,
+        "wordCount": len(article.text.split())
+    }
```
ssmall, line 115: This can be a method of the `Article` class

services/parser-database/parser.py
ssmall, line 119: Is there a way to combine lines 116 and 119 into a single method in `connector.py` that could then be more easily mocked? It may become easier to do this when the `dict` transformation is a method of `Article` (as suggested above)

services/parser-database/tests/test_articles.py
```diff
@@ -0,0 +1,36 @@
...
+    articles_to_send = articles.strip()
+    articles_to_send = articles_to_send.split()
+    recieved_articles = identify_articles(articles_to_send)
+    assert len(recieved_articles) == 6
+    assert recieved_articles[0].text == "los ciclistas se deben de orillar"
+    assert recieved_articles[1].text == "los ciclistas no se deben de orillar"
+    assert recieved_articles[2].text == "velocidad maxima de 10kmh"
+    assert recieved_articles[3].text == "carros de mas de 3/4 toneladas requieren licencia de chofer"
+    assert recieved_articles[4].text == "no te pueden infraccionar en navidad"
+    assert recieved_articles[5].text == "marcar al seguro y al 911 al chocar"
```
ssmall, line 36: The level of detail in this test is perfect! One suggestion: instead of calling `articles.strip()` and `articles_to_send.split()` in the test, why not just pre-process `articles` to be in the correct format?

services/parser-database/tests/test_endpoints.py
ssmall: Per our GVC chat, once the unit tests for the other pieces are in place, these tests can become a lot simpler by mocking out the dependencies rather than testing the entire application logic

services/parser-database/apiSpec.yaml
```diff
@@ -0,0 +1,134 @@
...
+    required:
+    - keywords
+  ArticleKeywords:
+    type: "object"
+    properties:
+      keyword:
+        type: "array"
+        items:
+          type: "object"
+          $ref: '#/definitions/ArticleKeywordFrequencyPair'
```
ssmall, line 105: Based on [this documentation](https://swagger.io/docs/specification/data-models/dictionaries/), I think what you want this to be is

```yaml
ArticleKeywords:
  type: object
  additionalProperties:
    type: object
    $ref: '#/definitions/ArticleKeywordFrequencyPair'
```

services/parser-database/apiSpec.yaml
```diff
@@ -0,0 +1,129 @@
...
+        type: "array"
+        items:
+          type: "string"
+    required:
+    - keywords
+  ArticleKeywords:
+    type: object
+    additionalProperties:
+      type: object
+      $ref: '#/definitions/ArticleKeywordFrequencyPair'
```
ssmall, line 102: Based on the discussion this morning, we'll update this to reflect the current implementation and then change the API implementation later. In that case, I believe this should be defined:

```yaml
ArticleKeywords:
    type: object
    additionalProperties:
      type: object
      additionalProperties:
        type: integer
```

PR #23, opened 2020-08-11
deployment (containerize app) Containerize frontend and backend services

oscardavidrm:
+ Added needed .dockerignore and Dockerfile into app/ folder to containerize the frontend component
+ Removed app/app.yaml and app/cloudbuild.yaml config files for App Engine since they're not longer necessary
+ Added needed .dockerignore and Dockerfile into server/ folder to containerize the backend component
+ Removed app/app.yaml config file for App Engine since its not longer necessary

+ Added axios library to handle HTTP communication with the search service component.
+ Added _mocked_ database array to search for articles with the articleIds returned from the search service component.


server/resolvers/article/queries.js
```diff
@@ -1,11 +1,14 @@
+import axios from 'axios'
+import {SEARCH_SERVICE_ENDPOINT} from "../../config"
```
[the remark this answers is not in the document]
oscardavidrm: Oh! Thanks! I know it didn't pass yesterday, but I wasn't able to find the error hahaha.

Thx! It's done now.

server/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:10
+
+# If PORT variable is not passed as argument during image build, it defaults to 4000
+# i.e. docker build --build-arg PORT=${ANY PORT} ...
+ARG PORT=4000
+ENV PORT=${PORT}
```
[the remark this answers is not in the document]
oscardavidrm: Yeah, if you wanna build locally, it's always good to have a default value to startup the service. That way, anyone who's running it doesn't run into problems.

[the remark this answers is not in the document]
jaimehisao: And if you want to run it with different values without rebuilding, can you change this with the `--env`flag at run?

app/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Step 0 - Build
+FROM tiangolo/node-frontend:10 AS build
+
+# If REACT_APP_HTTP_URI variable is not passed as argument during image build,
+# it defaults to "http://localhost:4000"
+# i.e. docker build --build-arg REACT_APP_HTTP_URI=${ANY REACT_APP_HTTP_URI} ...
+ARG REACT_APP_HTTP_URI="http://localhost:4000"
+ENV REACT_APP_HTTP_URI=$REACT_APP_HTTP_URI
```
ssmall, line 8: It probably doesn't make sense to treat this as a build arg, so you can just omit the `ARG` line here and set the default value directly on the `ENV` line

server/Dockerfile
ssmall, line 6: It probably doesn't make sense to treat this as a build arg, so you can just omit the `ARG` line here and set the default value directly on the `ENV` line

server/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+WORKDIR /app
+COPY package*.json ./
+RUN yarn install --production
+COPY . .
```
ssmall, line 11: Does this need to be three separate layers? Could it just be
```Dockerfile
COPY . .
RUN yarn install --production
```

server/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:10
+
+# If PORT variable is not passed as argument during image build, it defaults to 4000
+# i.e. docker build --build-arg PORT=${ANY PORT} ...
+ARG PORT=4000
+ENV PORT=${PORT}
+
```
ssmall, line 7: I believe you also need to specify an `ENV` var for `SEARCH_SERVICE_ENDPOINT`

server/resolvers/article/queries.js
```diff
@@ -24,67 +27,78 @@ const articleQueries = {
...
+    const articleIds = [...data.articles]
+
+    const articles = []
+    for (const id of articleIds) {
+      for (const article of database) {
+        if (id === article.id) { articles.push(article) }
+      }
+    }
+
+    return articles
```
ssmall, line 101: Are there unit tests for this?

app/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Step 0 - Build
+FROM tiangolo/node-frontend:10 AS build
```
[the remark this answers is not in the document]
oscardavidrm: I did start using the official node image, but I found this other one (which is based on the official as well) that comes with a preconfigured Nginx server to speed up the process. I didn't want to start raising issues with Nginx due to configuration and what not.

From `tiangolo/node-frontend` [docs](https://github.com/tiangolo/node-frontend):

> This Docker image simplifies the process of creating a full Node.js environment for frontend development with multistage building.

> It also includes a default Nginx configuration for your frontend application, so in multi-stage Docker builds you can just copy it to an Nginx "stage" and have an always freshly compiled production ready frontend Docker image for deployment.

So in essence, what it brings to the table, is a preconfigured **Nginx server**.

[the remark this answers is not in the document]
oscardavidrm: I could deff switch it up to the official Node image, I'd just have to go through the Nginx config as well.

Wdyt?

server/Dockerfile
oscardavidrm: Oh! Sure I guess.

I was watching a video on how to optimize Docker through caching by splitting your commands, but I am not sure if what I'm doing actually works hahaha.

I'll just join everything to simplify the Dockerfile :)

app/Dockerfile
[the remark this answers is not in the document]
ssmall: The [nginx.conf](https://github.com/tiangolo/node-frontend/blob/master/nginx.conf) is actually really simple, it probably makes sense to just include it in your PR directly and use the official image. Here's why I recommend that: [tiangolo/node-frontend](https://hub.docker.com/r/tiangolo/node-frontend) has not been updated in over a year. There have almost certainly been Node.js bugs and security vulnerabilities that have been identified and fixed during that time interval. The best way to make sure you are able to easily update your Node.js version is to use the official Docker image.

server/Dockerfile
ssmall: It's nice to optimize the Docker build sometimes, but in this case it probably makes the most sense to keep the Dockefile simple

server/resolvers/article/queries.js
Domene99: This is the mocked version of article by id db endpoint

server/resolvers/article/queries.js
```diff
@@ -1,11 +1,14 @@
+import axios from 'axios'
+import { SEARCH_SERVICE_ENDPOINT } from '../../config'
+
 const articleQueries = {
-  articles: (_, { search }) => {
-    return [
+  articles: async (_, { search }) => {
```
Domene99, line 5: When the db endpoint is available this should be the only thing in this file right?

server/resolvers/article/queries.js
ssmall: There is still application logic in this method that can be unit tested, even if the articles are mocked out. For example, it is making a `POST` request to the search endpoint.

app/Dockerfile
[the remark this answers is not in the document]
oscardavidrm: Okay, that makes sense. Will do.

server/Dockerfile
oscardavidrm: Yeah, that is true. I'll be adding the DB variable as well.

server/resolvers/article/queries.js
oscardavidrm: Exactly, I'm actually already switching from the mocked data to the actual request to the DB endpoint.

server/resolvers/article/queries.js
oscardavidrm: I'm pending to add unit tests for this endpoint. But to keep PR as concise as possible, I think is a good idea to have it as a next step on a following PR.

ssmall: Please add unit tests before merging. Code should not be merged to main without unit tests.

server/resolvers/article/queries.js
```diff
@@ -1,90 +1,28 @@
...
+      const articleIds = [...data.articles]
+      const articles = []
+
+      for (const id of articleIds) {
+        const article = await axios.get(`${DATABASE_ENDPOINT}/${id}`)
+
+        article.minutesToRead = parseInt(article.wordCount) / AVERAGE_WORDS_PER_MINUTE
+        delete article.wordCount
+
+        articles.push(article)
```
ssmall, line 19: Please add unit tests to this PR.

oscardavidrm: Already added unit tests. :)

server/tests/article.test.js
```diff
@@ -0,0 +1,167 @@
+import 'babel-polyfill'
+import path from 'path'
+import { ApolloServer, gql } from 'apollo-server'
+import { describe, expect, test } from '@jest/globals'
+import { importSchema } from 'graphql-import'
+import { createTestClient } from 'apollo-server-testing'
+import resolvers from './mock-resolvers'
```
ssmall, line 7: In order for this unit test to add value, it needs to import the real resolver from `../resolvers/articles/queries.js`. As discussed offline, let's pair program this on Monday to work through the issues with mocking axios.

server/tests/article.test.js
```diff
@@ -0,0 +1,167 @@
...
+  }
+]
+
+describe('Article Queries', () => {
+  const client = createTestClient(server)
+
+  tests.forEach(currentTest => {
+    const { name, query, variables, expected } = currentTest
+
+    test(`query: ${name}`, async () => {
```
ssmall, line 159: Great use of table-based testing!

app/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+FROM node:10 AS build
+
+# If REACT_APP_HTTP_URI variable is not passed as argument during image build,
+# it defaults to "http://localhost:4000"
+# i.e. docker build --build-arg REACT_APP_HTTP_URI=${ANY REACT_APP_HTTP_URI} ...
+ARG REACT_APP_HTTP_URI="http://localhost:4000"
+ENV REACT_APP_HTTP_URI=$REACT_APP_HTTP_URI
+
+COPY . .
+RUN yarn global add serve
```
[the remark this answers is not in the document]
oscardavidrm: I wasn't able to solve the issue that was popping with Cloud Run regarding the "missing error.log file". I went through all Stackoverflow to try and find something. I did see some people discussing about having similar issues but none with any resolution.

I did as well an official [GCP Nginx setup for Cloud Run](https://cloud.google.com/community/tutorials/deploy-react-nginx-cloud-run) and that didn't work as well.

And basically all the hassle and complications forced me to look for alternatives. [Serve](https://www.npmjs.com/package/serve) is a very simple web server for serving static files. I like it because it's pretty straight forward and gets the job done. Plus it worked great with Cloud Run the very first time I tried.

server/tests/article.test.js
oscardavidrm: :)

server/tests/article.test.js
```diff
@@ -0,0 +1,119 @@
...
+            query articles($search: String!) {
+                articles(search: $search) {
+                    number
+                    content
+                }
+            }
+        `,
+    variables: { search: 'Es obligatorio usar casco con bicicleta?' },
+    response: { data: { articles: ['1'] } },
+    expected: { articles: [articles[0]].map(({ number, content }) => ({ number, content })) }
```
Domene99, line 45: Is this test doing something from within your code. To me it just looks that it's testing the functionality of .map(), something you shouldn't be testing

server/tests/article.test.js
```diff
@@ -0,0 +1,119 @@
...
+                    number
+                    content
+                    keywords
+                    minutesToRead
+                }
+            }
+        `,
+    variables: { search: 'Es obligatorio usar casco con bicicleta?' },
+    response: { data: { articles: ['1'] } },
+    expected: { articles: [articles[0]] }
```
Domene99, line 31: Is there a reason why response is 1 and expected is 0? Correct me if I'm wrong but is it because of the difference of the indexing of the mocked search articles v.s. yours?

server/tests/article.test.js
```diff
@@ -0,0 +1,119 @@
...
+    query: gql`
+            query articles($search: String!) {
+                articles(search: $search) {
+                    id
+                }
+            }
+        `,
+    variables: { search: '' },
+    response: { data: { articles: ['20'] } },
+    expected: [new GraphQLError(JSON.stringify({ error: 'No article matches such ID', code: 404 }))]
```
Domene99, line 88: How is this handled on the frontend?

server/tests/endpoints.test.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+]
+
+describe('Endpoint tests', () => {
+  searchApiTests.forEach(currentTest => {
+    const { name, query, response, expected } = currentTest
+
+    test(`query: ${name}`, async () => {
+      axios.post.mockImplementation((url, data) => {
+        expect(url).toEqual(SEARCH_ENDPOINT)
```
Domene99, line 50: Is this the line that expects the call to the search endpoint?

server/tests/endpoints.test.js
```diff
@@ -0,0 +1,81 @@
...
+        return response
+      })
+      const results = await searchApi(query)
+      return expect(results).toEqual(expected)
+    })
+  })
+
+  databaseApiTests.forEach(currentTest => {
+    axios.get.mockImplementation(url => {
+      expect(url).toMatch(new RegExp(`^${DATABASE_ENDPOINT}/[0-9a-zA-Z\\-_]+$`))
```
Domene99, line 61: Why does the database endpoint have a regex condition?

server/resolvers/article/queries.js
```diff
@@ -1,90 +1,29 @@
...
       }
-    ]
+
+      return articles
+    }
+
+    return new Error(JSON.stringify(payload.error))
```
ssmall, line 26: nit: Consider inverting the `if (!payload.error)` condition so that it looks like

```javascript
if (payload.err) return new Error(JSON.stringify(payload.error))
```

I think the code will be easier to read that way.

server/tests/endpoints.test.js
oscardavidrm: This to ensure that the mocked endpoint responses are tested correctly since it is technically possible to basically pass anything to the endpoint that doesn't fit how the endpoint format should look like. Since the response is mocked, the tests are going to pass, nonetheless, whenever updates get deployed, it's possible to experience bugs. This helps enforce the expected input values.

server/tests/endpoints.test.js
oscardavidrm: Yes, its the `axios` way to mock a response.

server/tests/article.test.js
oscardavidrm: Although this updates are expected for next PR, I was thinking of showing a popup containing the error data.

server/tests/article.test.js
oscardavidrm: Yeah, `articles[0]` is just referencing the 1st article available in the test set, which is Article 1.

server/tests/article.test.js
oscardavidrm: Yees, it's testing to ensure that that only the selected attributes from the request get fetched. So basically, that test case makes sure that GraphQL is fetching the minimum requested data.

server/tests/article.test.js
```diff
@@ -0,0 +1,119 @@
...
+import { importSchema } from 'graphql-import'
+import { createTestClient } from 'apollo-server-testing'
+import resolvers from '../resolvers'
+import { databaseApi, searchApi } from '../endpoints'
+import { articles } from './mocks'
+
+const typeDefs = importSchema(path.join(__dirname, '../typedefs/index.graphql'))
+const server = new ApolloServer({ typeDefs, resolvers })
+
+const tests = [
```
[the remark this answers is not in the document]
oscardavidrm: On it! :)

server/tests/article.test.js
```diff
@@ -0,0 +1,119 @@
+import 'babel-polyfill'
```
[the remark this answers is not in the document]
oscardavidrm: Oh yeah, that sounds all right, but that file naming actually is based of `jest` testing library. The testing library looks for `*.test.*` documents to run the tests. I already tried changing to something like `article-test.js` but it won't work. I'll look into `jest` docs to see if there are any configuration workarounds.

server/tests/endpoints.test.js
```diff
@@ -0,0 +1,81 @@
...
+import axios from 'axios'
+import { describe, expect, jest, test } from '@jest/globals'
+import { databaseApi, searchApi } from '../endpoints'
+import { articles } from './mocks'
+import { DATABASE_ENDPOINT, SEARCH_ENDPOINT } from '../config'
+
+jest.mock('axios')
+
+const searchApiTests = [
+  {
```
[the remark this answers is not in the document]
oscardavidrm: Yeah, technically they are there to verify that the **mocked services** return the appropriate payload. And what you say makes sense as in if the external service changes its payload format, these unit tests won't actually break.

I added them as per @ssmall suggestion. Spencer, is there any other consideration as to why testing the mocked endpoints is important?

server/tests/article.test.js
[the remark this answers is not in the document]
anniefu: Hmm yeah, there are configuration work arounds, but strangely not that well documented. I'm fine with you leaving it in the default `jest` format. I'd just recommend changing the name to `query.test.js` then.

server/tests/endpoints.test.js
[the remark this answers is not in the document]
ssmall: > Do any of these tests run through any additional logic from your frontend/backend code? Or only mocks for external service? 

The logic that they test is very concise (because Javascript), but I think it's valuable -- it's testing that the `searchApi` function does actually attempt to make a `post` request to the appropriate URL. Without this test, `searchApi` could be redefined to just `return undefined` and all other unit tests would still pass.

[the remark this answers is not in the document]
anniefu: Ah, ok, that's fair. I think testing the URL is a reasonable test, but is better done with just one test case, and the `expected` field of the test case should be the URL expected in that case.

I would say `query` and `response` fields can be dropped from the `searchApiTests` elements, and just choose a fixed one in the test's for-loop. The "unit" test for this is the endpoint routing, not the search engine service, so you shouldn't have to vary the service request/responses. I would expect the querying logic unit to be tested in a different test suite (which is done in the other test file).

server/tests/article.test.js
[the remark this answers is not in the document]
oscardavidrm: Okay :)

server/tests/endpoints.test.js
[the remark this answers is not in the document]
oscardavidrm: I just removed the aforementioned fields from the test cases @anniefu. Although, I am not completely sure if I got it right hahaha.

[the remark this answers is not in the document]
anniefu: Yep, that's the idea! Could you do the same for the databaseApiTests too, please?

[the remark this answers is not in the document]
oscardavidrm: For sure 👍

server/tests/article.test.js
Domene99: I think this specific error has no need to be displayed, if a result id matches no article just don't display that ID

Issue #27, opened 2020-08-14
Uncaught errors in keyword response causes confusing error message


ssmall:
The following code should check if the response from the keywords service is an error before trying to extract the `lan` property:

https://github.com/googleinterns/major-tom/blob/8cf9d3f2ee084856eac1516d9de9dbbd3fffec58/services/search/search_engine.py#L41-L44

Right now, it throws an unintuitive error when calling this service:
```
Traceback (most recent call last):
File "/app/main.py", line 22, in search_service
score_per_article = search.query(query)
File "/app/search_engine.py", line 43, in query
lan = response[\'lan\']
KeyError: \'lan\' 
```

but what is really happening is the keywords endpoint is returning an error.

PR #29, opened 2020-08-14
Raise keywords request as error exception. Make search_query method p…

Domene99:
…ublic for testing

closing #27

### DIFF
--- a/services/search/search_engine.py
+++ b/services/search/search_engine.py
@@ -41,6 +41,9 @@ class SearchEngine:
         response = requests.post(keywords_endpoint, json=query_text)
         response = response.json()
 
+        if 'error' in response:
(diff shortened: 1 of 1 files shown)


services/search/search_engine.py
```diff
@@ -71,7 +74,7 @@ def _calculate_score(self, frequency, weight, words, target_dict):
                     if score != 0:
                         target_dict[key] = score
 
-    def _search_query(self, keywords, synonyms):
+    def search_query(self, keywords, synonyms):
```
ssmall, line 77: Is this change related to the error exception somehow? It's not clear why it was included.

Domene99: This is not related at all, it is because pytest was not finding the method and since this was a small pr I decided to add it here.

PR #30, opened 2020-08-14
Add concurrent api calling for synonyms

Domene99:
closing #24


services/search/utils.py
```diff
@@ -11,15 +11,35 @@ def create_synonym_list_esp(word_arr, max_synonyms=5):
...
+    Create asynchronous requests list using the futures session
+    Attributes:
+        word_arr: list of all requests to be made
+        session: requests_futures instanced session
+    Return:
+        Asynchronous HTTP request list
+    """
+    reqs = []
+    for word in word_arr:
+        reqs.append(session.get(constants.SPANISH_API_URL+word))  # pylint: disable=no-member
```
oscardavidrm, line 43: Does the `Futures` library has a limit for concurrent requests that we should care about? Imagine if you go through a `word_arr` containing 1,000,000 words, would the process run out of memory? Or could it handle it?

services/search/utils.py
```diff
@@ -11,15 +11,35 @@ def create_synonym_list_esp(word_arr, max_synonyms=5):
...
+
         if len(word_synonyms) > max_synonyms:
             word_synonyms = word_synonyms[:max_synonyms]
         for synonym in word_synonyms:
             synonyms.append(synonym['sinonimo'])
 
     return synonyms
+
+
+def create_conc_reqs(word_arr, session):
```
ssmall, line 32: I think this would be a good time to add unit tests for the methods in the file. Looking at what's in `main`, it looks like this is untested right now.

services/search/utils.py
```diff
@@ -11,15 +11,35 @@ def create_synonym_list_esp(word_arr, max_synonyms=5):
...
     synonyms = []
-    for word in word_arr:
-        payload = requests.get(constants.SPANISH_API_URL+word)
-        payload.raise_for_status()
 
-        word_synonyms = payload.json()['sinonimos']
+    for req in reqs:
+        resp = req.result()
+
+        word_synonyms = resp.json()['sinonimos']
```
ssmall, line 22: What if the response contains an error?

services/search/utils.py
Domene99: I think there's not a limit. There's a max workers limit set in the constructor which specifies the number of threads, I believe if we exceed it we will just do a 2nd, 3rd or nth cycle to fulfill all our requests. Having said that do you believe we should set a hard limit of the input query on the front end? I think a user shouldn't even exceed even a 100 word search query let alone a million one.

services/search/testing/test_search_engine.py
```diff
@@ -1,5 +1,70 @@
+import mock  # pylint: disable=import-error
```
jaimehisao, line 1: nit: if you do `from unitttest import mock` there is no need to disable the error

services/search/utils.py
```diff
@@ -12,11 +42,17 @@ def create_synonym_list_esp(word_arr, max_synonyms=5):
...
-        payload.raise_for_status()
 
-        word_synonyms = payload.json()['sinonimos']
+    synonym_extractor = SynonymExtractor(max_synonyms)
+
+    for req in synonym_extractor.create_conc_reqs(word_arr):
+        resp = synonym_extractor.req_to_json(req)
+
+        if 'sinonimos' not in resp:
+            raise Exception("Unexpected synonyms API response format")
```
jaimehisao, line 52: Would a logging make sense here too?

services/search/testing/test_search_engine.py
Domene99: Really weird since they should be the same according to https://pypi.org/project/mock/

jaimehisao: If you are using the `mock` from `pip` that is why you are getting an error, since it is an external library. the one I suggested is internal and that is why it doesn't require the error suppresion. Any reason why you are not using the integrated `unittest`?

services/search/utils.py
```diff
@@ -12,11 +43,18 @@ def create_synonym_list_esp(word_arr, max_synonyms=5):
...
-        payload.raise_for_status()
 
-        word_synonyms = payload.json()['sinonimos']
+    synonym_extractor = SynonymExtractor(max_synonyms)
+
+    for req in synonym_extractor.create_conc_reqs(word_arr):
+        resp = synonym_extractor.req_to_json(req)
+        logging.info(resp)
+
+        if 'sinonimos' not in resp:
```
oscardavidrm, line 53: Nice error handling!

services/search/testing/test_search_engine.py
```diff
@@ -1,5 +1,70 @@
+from unittest import mock
```
oscardavidrm, line 1: I like the testing, it is pretty straight forward.

services/search/utils.py
```diff
@@ -1,7 +1,38 @@
...
 
+class SynonymExtractor:
+    """
+    Wrapper class for requests_futures methods
+    """
+    def __init__(self, max_synonyms):
+        self.max_synonyms = max_synonyms
+        self.session = FuturesSession()
+
+    def create_conc_reqs(self, word_arr):
```
ssmall, line 14: nit: it's an implementation detail that the requests are concurrent, so it would be more appropriate to name this method `create_reqs`

services/search/utils.py
```diff
@@ -1,7 +1,38 @@
...
+        for word in word_arr:
+            reqs.append(self.session.get(constants.SPANISH_API_URL+word))  # pylint: disable=no-member
+
+        return reqs
+
+    def req_to_json(self, json):
+        """
+        Returns json from request
+        """
+        return json.result().json()
```
ssmall, line 33: nit: The naming here is a little confusing. This method is actually getting the response and converting it to JSON, so maybe a better name would be something like `get_response_json`. 

Also, the first arg is called `json` but it's actually a `request`. Consider giving it a less confusing name.

services/search/utils.py
```diff
@@ -1,7 +1,38 @@
-import requests  # pylint: disable=import-error
+import logging
```
anniefu, line 1: [nit] I'd rename htis file to `synonym_extractor.py` now

services/search/utils.py
```diff
@@ -1,7 +1,38 @@
...
+        create an asynchronous request list using requests-futures
+        Attributes:
+            word_arr: A list of words which each is a different request
+            session: A requests-futures session
+        Return:
+            A list of requests for every word
+        """
+        reqs = []
+        for word in word_arr:
+            reqs.append(self.session.get(constants.SPANISH_API_URL+word))  # pylint: disable=no-member
```
anniefu, line 25: I'd recommend calling `req_to_json()` on the return of `self.session.get` before appending it to reqs. That way you can keep the function private to the class & avoid making the caller call `req_to_json` too.

Also, I think`reqs` here is really `responses`.

services/search/testing/test_search_engine.py
```diff
@@ -1,5 +1,70 @@
...
+
+
+def test_one_synonym():
+    """
+    Tests one synonym call with default max synonyms (5)
+    """
+    result = ["my_synonym"] * constants.DEFAULT_MAX_SYNONYMS
+
+    with mock.patch('utils.SynonymExtractor.create_conc_reqs', side_effect=reqs_list):
+        with mock.patch('utils.SynonymExtractor.req_to_json', side_effect=to_json):
```
anniefu, line 28: Hmm, I would have expected just one mock for `self.session.get()`. I think if you mock out that response, you don't need to mock out anything else in `SynonymExtractor`.

Domene99: I thought so too and it should be possible, but the response object seems to be different than a normal response, where a normal response can be accessed using only `.json()` v.s. here you need to use `.result().json()` I did spend some time looking for the response object in the requests_futures repo and couldn't find it, this issue was discussed with @ssmall on our 1:1 coding session and this is the solution he proposed to me.

Domene99: But with your other suggestion this is possible

services/search/utils.py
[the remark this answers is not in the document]
Domene99: Learned from the best

oscardavidrm:
Remember to resolve the active conflicts you have!

Domene99:
> Remember to resolve the active conflicts you have!

Will do, but I've decided on doing this just before merging, because if I resolve now and then I have some comments to attend to, my local will conflict with the merges made here on github.

services/search/synonym_extractor.py
```diff
@@ -0,0 +1,63 @@
...
+            A list of requests for every word
+        """
+        responses = []
+        for word in word_arr:
+            responses.append(self.session.get(constants.SPANISH_API_URL+word))  # pylint: disable=no-member
+
+        responses_json = [self.get_response_json(resp) for resp in responses]
+        return responses_json
+
+    def get_response_json(self, resp):
```
anniefu, line 30: [nit] I believe you can make this private to the class now by renaming it `__get_response_json`: https://www.tutorialspoint.com/What-does-double-underscore-prefix-do-in-Python-variables#:~:text=If%20your%20class%20is%20intended,mangled%20into%20the%20attribute%20name.

services/search/testing/test_search_engine.py
anniefu: I see, there's technically like two for-loops going untested here, which isn't ideal.

Because python is weakly typed and just keeps trying to execute lines until it gets an error, I believe you can actually use a mock class defined in just the test file to simulate an object with `.result().json()` and return that as the mock of `self.session.get()`. It would take two nested classes though.

Something like:
```python
class MockJson:
    __init__(self, json_result_value_you_want_to_use_for_test):
         self.json = json_result_value_you_want_to_use_for_test
    def json(self):
         return self.json

class MockResult:
    __init__(self, json_result_value_you_want_to_use_for_test):
         self.mock_json = new MockJson(json_result_value_you_want_to_use_for_test)
    def result(self):
        return self.mock_json

def test():
    req_list = [new MockResult('{"some": "json"}'), new MockResult('{"some": "json"})]
    with mock.patch('self.session.get', side_effect=reqs_list)
```

Issue #35, opened 2020-08-14
Add debug logging for search service


ssmall:
Some ideas for things to log in the service:

- content of each requests
- any errors
- content of response

This list is just a starting point, there are probably other things that will be helpful to log

Issue #34, opened 2020-08-14
Add debug logging for keywords service


ssmall:
Some ideas for things to log in the service:

- content of each requests
- any errors
- content of response

This list is just a starting point, there are probably other things that will be helpful to log

PR #37, opened 2020-08-17
Add logging to keywords and search

Domene99:
Logging missing in search/utils.py to avoid merge conflicts with existing #30 

Closing #34 
Closing #35

### DIFF
--- a/services/keywords/extract.py
+++ b/services/keywords/extract.py
@@ -34,7 +34,7 @@ def extract_keywords(text):
         the lemma and part of speech for said keywords
     """
     gcloud_response = gcloud_syntax_extraction(text)
-    logging.info("gcloud syntax response:", gcloud_response)  # pylint: disable=logging-too-many-args
+    logging.info("gcloud syntax response: %s", gcloud_response)
 
     tokens_shortened = []
     for token in gcloud_response.tokens:
--- a/services/keywords/extract.py
+++ b/services/keywords/extract.py
@@ -34,8 +34,8 @@ def extract_keywords(text):
         the lemma and part of speech for said keywords
     """
     gcloud_response = gcloud_syntax_extraction(text)
-    logging.info("gcloud syntax response:", gcloud_response)
-    
+    logging.info("gcloud syntax response:", gcloud_response)  # pylint: disable=logging-too-many-args
+
     tokens_shortened = []
     for token in gcloud_response.tokens:
         part_of_speech_tag = enums.PartOfSpeech.Tag(token.part_of_speech.tag).name
@@ -45,5 +45,5 @@ def extract_keywords(text):
             tokens_shortened.append(token_data)
 
     response = {'lan': gcloud_response.language, 'tokens': tokens_shortened}
-    
+
     return response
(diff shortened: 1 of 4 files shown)


Domene99:
Due to the fact that our pylint version doesn't seem to have this bug fix, I'm disabling this error (logging-too-many-args) manually
https://github.com/PyCQA/pylint/issues/3362

services/keywords/extract.py
```diff
@@ -33,6 +34,7 @@ def extract_keywords(text):
         the lemma and part of speech for said keywords
     """
     gcloud_response = gcloud_syntax_extraction(text)
+    logging.info("gcloud syntax response:", gcloud_response)  # pylint: disable=logging-too-many-args
```
ssmall, line 37: You're getting a linter error because this is an incorrect usage of `logging.info`. You need to specify a format string: https://docs.python.org/3.8/library/logging.html#logging.debug

So this should be:
```python
logging.info("gcloud syntax response: %s", gcloud_response)
```

services/keywords/main.py
```diff
@@ -11,13 +12,21 @@ def get_keywords_service(request):
         json of language detected and keywords for text
     """
     json = request.get_json()
+    logging.info("Endpoint request:", json)  # pylint: disable=logging-too-many-args
+
     if json is None or "text" not in json:
-        return {"error": {"message": "ValueError: Expected 'text' field in json body is missing"}}
+        error = {"error": {"message": "ValueError: Expected 'text' field in json body is missing"}}
+        logging.error(error['error']['message'])  # pylint: disable=logging-too-many-args
```
ssmall, line 19: This would be more readable if you made `message` a variable and used it in both places.

services/keywords/main.py
```diff
@@ -11,13 +12,21 @@ def get_keywords_service(request):
...
 
     text = json["text"]
     try:
         response = extract.extract_keywords(text)
     except Exception as e:
-        return {"error": {"message": getattr(e, 'message', str(e)),
-                          "trace": traceback.format_exc()}}
+        error = {"error": {"message": getattr(e, 'message', str(e)),
+                           "trace": traceback.format_exc()}}
+        logging.error(error['error'])  # pylint: disable=logging-too-many-args
```
ssmall, line 28: This would be more readable if you made `message` a variable and used it in both places.

services/keywords/extract.py
Domene99: Wow that's a pretty limited selection of string interploration. Not even f string interploration works with pylint

services/search/search_engine.py
```diff
@@ -38,8 +39,12 @@ def query(self, query):
         """
         query_text = {'text': query}
         keywords_endpoint = os.getenv('KEYWORDS_ENDPOINT') is not None or 'http://localhost:8081'
```
jaimehisao, line 41: lol this is helpful for my part

Domene99: Cool! hahahah

services/keywords/main.py
```diff
@@ -11,13 +12,22 @@ def get_keywords_service(request):
         json of language detected and keywords for text
     """
     json = request.get_json()
+    logging.info("Endpoint request: %s", json)
```
oscardavidrm, line 15: 🙏

PR #38, opened 2020-08-18
Integrate db endpoint

Domene99:
Integrate into db endpoint and mock result in unit tests
Closing #39


services/search/testing/test_search_engine.py
```diff
@@ -19,16 +21,17 @@ def test_double_weights_synonyms():
...
+
+def test_multiple_articles():
+    """
+    Have the query apply for two articles
+    """
+    result = {3: 2, 4: 2}
+
+    with mock.patch('requests.post', return_value=constants.KEYWORDS_DB_MULTIPLE):
+        assert result == SearchEngine().search_query(constants.KEYWORDS_MULTIPLE,
+                                                     constants.SYNONYMS_MULTIPLE)
```
jaimehisao, line 37: Good test, as long as this data is combined into a single array, it should be fine on my end.

services/search/testing/test_constants.py
```diff
@@ -1,9 +1,16 @@
...
+                      "emplear": {}, "utilizar": {1: 1}, "gastar": {}, "acostumbrar": {},
+                      "practicar": {},
                       "copa": {}, "suelo": {}, "pezuna": {}, "vaso": {}, "cabeza": {}}
 
+KEYWORDS_DB_MULTIPLE = {"taxi": {3: 2}, "tener": {}, "calcomania": {4: 1}, "placa": {4: 1},
+                        "taximetro": {},
+                        "haber": {}, "poseer": {}, "detentar": {}, "sujetar": {}, "coger": {},
+                        "copia": {}, "reproduccion": {}, "estampa": {}, "adhesivo": {},
+                        "lamina": {}, "plancha": {}, "chapa": {}, "hoja": {}, "pelicula": {}}
+
```
jaimehisao, line 13: the return looks right, a dictionary of dictionaries.

services/search/testing/test_search_engine.py
Domene99: This is the search endpoint, only oscar should be calling it

services/search/testing/test_constants.py
Domene99: Cool thanks!

services/search/testing/test_search_engine.py
```diff
@@ -9,8 +10,9 @@ def test_same_weights_for_all():
     """
     result = {1: 3}
 
-    assert result == SearchEngine().search_query(constants.KEYWORDS_ARTICLE_1,
-                                                 constants.SYNONYMS_ARTICLE_1)
+    with mock.patch('requests.post', return_value=constants.KEYWORDS_DB_MOCK_1):
+        assert result == SearchEngine().search_query(constants.KEYWORDS_ARTICLE_1,
+                                                     constants.SYNONYMS_ARTICLE_1)
```
ssmall, line 15: It would be good to also use [`assert_called_once_with`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once_with) to verify that `requests.post` was called with the correct arguments. Now that you're mocking it, your code could call it with random arguments and this test would technically still pass.

services/search/search_engine.py
```diff
@@ -67,7 +66,14 @@ def query(self, query):
 
         return self.search_query(keywords, synonyms)
 
-    def _calculate_score(self, frequency, weight, words, target_dict):
+    def _calculate_individual(self, word, frequency, weight, target_dict):
+        if word in frequency:
+            for key, value in frequency[word].items():
+                score = target_dict.get(key, 0) + value * weight
```
oscardavidrm, line 72: A+ in math.

services/search/search_engine.py
```diff
@@ -94,20 +99,11 @@ def search_query(self, keywords, synonyms):
...
 
-        # keywords_json = {"keywords": keywords}
-        # synonyms_json = {"keywords": synonyms}
-        # article_keywords_frequency = requests.post(db_endpoint, json=keywords_json)
-        # article_synonyms_frequency = requests.post(db_endpoint, json=synonyms_json)
-
-        article_keywords_frequency = testing.test_constants.KEYWORDS_DB_MOCK_1
-        article_synonyms_frequency = testing.test_constants.SYNONYMS_DB_MOCK_1
+        keywords_json = {"keywords": keywords+synonyms}
+        article_keywords_frequency = requests.post(os.getenv('DB_ENDPOINT'), json=keywords_json)
```
oscardavidrm, line 103: Are you planning to do any error handling for the `DB_ENDPOINT`??

Domene99: Talked with @jaimehisao and he will respond with errors in the same format as me

services/search/testing/test_search_engine.py
```diff
@@ -9,8 +11,12 @@ def test_same_weights_for_all():
     """
     result = {1: 3}
 
-    assert result == SearchEngine().search_query(constants.KEYWORDS_ARTICLE_1,
-                                                 constants.SYNONYMS_ARTICLE_1)
+    with mock.patch('requests.post', return_value=constants.KEYWORDS_DB_MOCK_1) as mocked:
+        assert result == SearchEngine().search_query(constants.KEYWORDS_ARTICLE_1,
+                                                     constants.SYNONYMS_ARTICLE_1)
+        mocked.assert_called_once_with(
+            os.getenv('DB_ENDPOINT'),
```
ssmall, line 18: I don't think this makes sense in a test. How will `DB_ENDPOINT` get set for the test? I expect that if I run `pytest test_search_engine.py` it should just work, but including this in your unit test means that I have to set an env variable every time I run unit tests or else the tests will fail.

This means you probably need to refactor `search_query` to not call `os.getenv` directly but rather to import the value from somewhere, and then you can mock that import for your unit tests.

services/search/testing/test_search_engine.py
```diff
@@ -9,8 +11,13 @@ def test_same_weights_for_all():
     """
     result = {1: 3}
 
-    assert result == SearchEngine().search_query(constants.KEYWORDS_ARTICLE_1,
-                                                 constants.SYNONYMS_ARTICLE_1)
+    with mock.patch('requests.post', return_value=constants.KEYWORDS_DB_MOCK_1) as mocked:
+        with mock.patch('env.get_db_endpoint', return_value="http://somerealurl.com"):
```
ssmall, line 15: it would make the test cleaner to extract the mock URL to a variable and reuse it in the assert

services/search/testing/test_search_engine.py
```diff
@@ -19,16 +26,25 @@ def test_double_weights_synonyms():
...
-#
-# def test_multiple_articles():
-#    """
-#    Have the query apply for two articles
-#    """
-#    result = {3: 2, 4: 2}
-#    assert result == SearchEngine(constants.ARTICLES).search_query(constants.KEYWORDS_MULTIPLE,
-#                                                                   constants.SYNONYMS_MULTIPLE)
+    with mock.patch('requests.post', return_value=constants.KEYWORDS_DB_MOCK_1) as mocked:
+        with mock.patch('env.get_db_endpoint', return_value="http://somerealurl.com"):
```
ssmall, line 30: it would make the test cleaner to extract the mock URL to a variable and reuse it in the assert

services/search/testing/test_search_engine.py
```diff
@@ -19,16 +26,25 @@ def test_double_weights_synonyms():
...
+
+
+def test_multiple_articles():
+    """
+    Have the query apply for two articles
+    """
+    result = {3: 2, 4: 2}
+
+    with mock.patch('requests.post', return_value=constants.KEYWORDS_DB_MULTIPLE) as mocked:
+        with mock.patch('env.get_db_endpoint', return_value="http://somerealurl.com"):
```
ssmall, line 45: it would make the test cleaner to extract the mock URL to a variable and reuse it in the assert

PR #40, opened 2020-08-18
Connection with NLP Service

jaimehisao:
This PR connects the Parser with the NLP service.


jaimehisao:
Looking for review, specially from @Domene99  since I call his endpoint here.

ssmall:
This PR currently includes all the changes from #22 since that PR is not merged yet. Please mark this PR as a Draft until #22 is merged.

services/parser-database/connector.py
```diff
@@ -0,0 +1,70 @@
...
+        text (sting): text to extract keywords from
+
+    Returns:
+        [list]: list of extracted keywords
+    """
+
+    extracted_keywords = []
+    request = {'text': text}
+    nlp_output = requests.post(os.getenv("KEYWORDS_SERVICE"), json=request)
+    # nlp_output = keywordmock.get_keywords(text)
```
oscardavidrm, line 41: nit: remove comment.

services/parser-database/connector.py
```diff
@@ -0,0 +1,70 @@
...
+    json_output = nlp_output.json()
```
oscardavidrm, line 42: It seems that this would break if the response is not in JSON format (i.e., Javier's service returns an error as response). Probably it would be worth it to wrap this in a `try/except` block.

services/parser-database/connector.py
```diff
@@ -0,0 +1,70 @@
...
+    for keyword in json_output["tokens"]:
+        extracted_keywords.append(keyword["lemma"])
+    return extracted_keywords
+
+
+def get_article_by_number(art_num):
+    if art_num in articles_in_memory:
```
oscardavidrm, line 49: I am not really sure how does the `in` keyword works in Python, but it seems that it traverses over all object attributes to find the article.

Check this [link](https://stackoverflow.com/a/610893/11672866) out. They use `hasattr()` method to find the attribute. I don't know if it's faster, but you should look at it.

services/parser-database/connector.py
```diff
@@ -0,0 +1,70 @@
...
+    Args:
+        keywords_list (list): Keyword(s) to look for
+
+    Returns:
+        list: articles that match such keyword(s)
+    """
+    matching_articles = {}
+    for keyword in keywords_list:
+        articles_that_match_keyword = {}
+        if keyword in keywords_in_memory:
```
oscardavidrm, line 66: Same as comment above 👁️

services/parser-database/connector.py
Domene99: Avg: O(1) https://stackoverflow.com/questions/13884177/complexity-of-in-operator-in-python

anniefu:
Please rebase the PR on master, I would expect this to be a smaller change than what it is now, since none of the database functions should need to change.

services/parser-database/connector.py
```diff
@@ -0,0 +1,70 @@
...
+    Args:
+        text (sting): text to extract keywords from
+
+    Returns:
+        [list]: list of extracted keywords
+    """
+
+    extracted_keywords = []
+    request = {'text': text}
+    nlp_output = requests.post(os.getenv("KEYWORDS_SERVICE"), json=request)
```
Domene99, line 40: You might want to see if I ever return you an {'error'}, since your code would crash in L43

jaimehisao: done!

Issue #36, opened 2020-08-14
Single word queries are not supported


ssmall:
When I try to search for "bicicleta", I get the following error:

```
Traceback (most recent call last):
File "/app/main.py", line 22, in search_service
score_per_article = search.query(query)
File "/app/search_engine.py", line 46, in query
raise ValueError(f\'{lan} not supported\')
ValueError: en not supported
```

It seems that the search engine is assuming that my query is English.

PR #41, opened 2020-08-18
Remove exception check for languages not supported

Domene99:
Changed to log as warning
Closing #36

### DIFF
--- a/services/search/search_engine.py
+++ b/services/search/search_engine.py
@@ -52,7 +52,7 @@ class SearchEngine:
         lan = response['lan']
 
         if lan not in constants.SUPPORTED_LANGUAGES:
-            raise ValueError(f'{lan} not supported')
+            logging.warning("%s not supported", lan)
 
         keywords = []
 


services/search/search_engine.py
```diff
@@ -52,7 +52,7 @@ def query(self, query):
         lan = response['lan']
 
         if lan not in constants.SUPPORTED_LANGUAGES:
-            raise ValueError(f'{lan} not supported')
+            logging.warning("%s not supported", lan)
```
ssmall, line 55: It would be cool if you and @oscardavidrm could figure out a way to include this warning in the response so that it could still be displayed on the frontend without failing the request.

In other words, still do the search as normal and return any results, but if the results are empty and the warning is present, the frontend could display an informational popup about the language.

Domene99: That sounds like something cool, I'll have to think on how to raise the warning without interrupting tho.

oscardavidrm: Yeah, that sounds interesting. If you raise an error, it is pretty straight forward since the frontend will get the same format we agreed upon and it would basically just be displaying it in the popup. 

As per my understanding, this has come with some issues right? If you cannot raise `ValueError`, we could probably look into an additional attribute in your response, a `warnings array` for example containing strings with relevant information (such as the one you're pointing out) and I could look for that additional attribute and group those so called warnings.

shaargtz:
LGTM!

PR #45, opened 2020-08-19
feature (loading) Add loading functionality when searching.

### LINKED ISSUE #46: Add loading visual feature to frontend

oscardavidrm:
Added a progress bar component inside search-bar/index.js to visually show when the request has been sent and is waiting for a response.

Closes #46

### DIFF
--- a/app/src/views/search-engine/components/search-bar/index.js
+++ b/app/src/views/search-engine/components/search-bar/index.js
@@ -1,9 +1,9 @@
 import React from 'react'
-import { AppBar, Toolbar, Avatar, Input, IconButton } from '@material-ui/core'
+import { AppBar, Toolbar, Avatar, Input, IconButton, Fade, LinearProgress } from '@material-ui/core'
 import { Search as SearchIcon } from '@material-ui/icons'
 import styles from './styles'
 
-const SearchBar = ({ setSearch, refetch }) => {
+const SearchBar = ({ setSearch, refetch, setLoading, loading }) => {
   const classes = styles()
 
   return (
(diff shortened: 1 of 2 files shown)


app/src/views/search-engine/components/search-bar/index.js
```diff
@@ -25,11 +25,23 @@ const SearchBar = ({ setSearch, refetch }) => {
               root: classes.inputRoot,
               input: classes.inputInput
             }}
-            onChange={({ target: { value } }) => setSearch(value)}
+            onChange={({ target: { value } }) => {
+              if (value?.length !== 0) {
+                setLoading(true)
+                setSearch(value)
```
Domene99, line 31: What happens if you setSearch when value?.length == 0? I think it would be nicer code to have this as
`setLoading(value?.length !== 0)`
`setSearch(value)`
or possibly validate the condition in setSearch

oscardavidrm: Yes, it actually does look nicer, but if you notice, I am trying to avoid calling `setSearch()` if the value is an empty string. Whenever `setSearch()` gets invoked, a network request get queued, and we want to avoid that if `value` is empty.

app/src/views/search-engine/components/search-bar/index.js
```diff
@@ -25,11 +25,23 @@ const SearchBar = ({ setSearch, refetch }) => {
...
+              } else {
+                setLoading(false)
+              }
+            }}
```
ssmall, line 35: You're probably not gonna like this question but ... how hard would it be to unit test this?

oscardavidrm: 😆 

I mean, it's totally possible. I don't think it would be "hard" per se, but of course I'd have to research a little bit to understand how to do it. Rather than hard, there is an implicit learning curve in there that naturally will take time 👁️ .

I was hoping to focus on the caching path expansion, but I could definitely prioritize this if needed.

Wdyt?

ssmall: It would be good to have, but I don't think it's more important than other improvements you could make. I'll leave it up to you to whether you want to figure out how to unit test this now or come back to it later if there's time.

oscardavidrm: Yes, that makes sense, I think I'd like to focus on getting between today/tomorrow the caching service running, and once that gets done (and I wouldn't expect it to be THAT complex), I will definitely focus on getting the React components tested.

PR #51, opened 2020-08-20
PR that connects the Parser and NLP

jaimehisao:
NLP and Parser are officially connected!


services/parser-database/connector.py
```diff
@@ -20,20 +21,23 @@ def get_documents_to_parse():
 
 
 def get_keywords(text):
-    """Get keywords that relate to this article
+    """Get keywords that relate to this article (Javier's service)
```
anniefu, line 24: Javier, exposed. (But actually it would be better to refer to this as the keyword extraction service).

ssmall, line 24: This comment probably is not appropriate. It should be referred to as the "keywords service"

services/parser-database/connector.py
```diff
@@ -20,20 +21,23 @@ def get_documents_to_parse():
...
 
     Args:
         text (sting): text to extract keywords from
 
     Returns:
         [list]: list of extracted keywords
     """
-    splited_text = text.split()
```
anniefu, line 31: [nit] the past tense of split is actually just split, so this can just be "split_text"

services/parser-database/connector.py
```diff
@@ -20,20 +21,23 @@ def get_documents_to_parse():
...
-    return keywords
+
+    extracted_keywords = []
+    request = {'text': text}
+    nlp_output = requests.post(os.getenv("KEYWORDS_SERVICE"), json=request)
+    # nlp_output = keywordmock.get_keywords(text)
+    json_output = nlp_output.json()
+    for keyword in json_output["tokens"]:
+        extracted_keywords.append(keyword["lemma"])
+    return extracted_keywords
```
ssmall, line 40: there should be some corresponding changes to unit tests, right?

services/parser-database/connector.py
```diff
@@ -20,20 +21,23 @@ def get_documents_to_parse():
...
     Returns:
         [list]: list of extracted keywords
     """
-    splited_text = text.split()
-    keywords = [
-        splited_text[random.randint(0, len(splited_text) - 1)],
-        splited_text[random.randint(0, len(splited_text) - 1)],
-    ]
-    return keywords
+
```
anniefu, line 32: You probably want to do a try/catch block around this whole block so if there are errors contacting the service or the response is missing important fields, you can error handle

services/parser-database/connector.py
```diff
@@ -20,20 +21,23 @@ def get_documents_to_parse():
 
 
 def get_keywords(text):
```
anniefu, line 23: Is there a test for this function already?

services/parser-database/connector.py
jaimehisao: Yes, fixing!

jaimehisao: again haha, yes, I wen't ahead and changed that

services/parser-database/connector.py
jaimehisao: Yes!

services/parser-database/connector.py
```diff
@@ -20,20 +19,24 @@ def get_documents_to_parse():
...
-    return keywords
+    extracted_keywords = []
+    request = {'text': text}
+    nlp_output = requests.post(os.getenv("KEYWORDS_SERVICE"), json=request)
+    json_output = nlp_output.json()
+    if 'error' in json_output:
+        raise Exception(json_output['error']['message'])
+    for keyword in json_output["tokens"]:
+        extracted_keywords.append(keyword["lemma"])
+        print(extracted_keywords)
```
Domene99, line 38: logging?

oscardavidrm: +1

That doesn't display anything on the container logs right?

jaimehisao: Oops

services/parser-database/tests/test_articles.py~HEAD
```diff
@@ -0,0 +1,41 @@
+"""This tests only test (duh) the parser's ability of
```
anniefu, line 1: What's up with this file name?

PR #52, opened 2020-08-20
Improve search engine with 3 new factors

Domene99:
Return the term frequency inverse document frequency in the db endpoint, [tfidf](http://www.tfidf.com/) or [wikipedia](https://en.wikipedia.org/wiki/Tf%E2%80%93idf), previous version of our algorithm only took into consideration term frequency, aka incidences. The 3 new factors are the density of the article, the total number of documents, and the number of documents with the term in it.
Closing #53


Domene99:
Tested locally, unittesting should come soon.

services/parser-database/connector.py
```diff
@@ -61,6 +62,33 @@ def get_articles_that_match_keywords(keywords_list):
...
+    matching_articles = {}
+    for keyword in keywords_list:
+        articles_that_match_keyword = {}
+        if keyword in keywords_in_memory:
+            for article in keywords_in_memory[keyword]:
+                # tfidf computation
+                term_density_in_article = article["frequency"]/article["wordCount"]
+                document_frequency = len(articles_in_memory)/len(keywords_in_memory[keyword])
+                inverse_doc_freq = np.log(document_frequency)
+                value = term_density_in_article * inverse_doc_freq
```
oscardavidrm, line 85: **is dis advanced math?**
isdis

services/parser-database/connector.py
```diff
@@ -73,8 +101,10 @@ def save_keywords_in_memory(keywords, article):
         if keyword not in keywords_in_memory:
             keywords_in_memory[keyword] = []
         keywords_in_memory[keyword].append({
+            "id": article["id"],
             "articleNumber": article["articleNumber"],
```
oscardavidrm, line 105: nit: Update all references of `articleNumber` to `number` since that was the one of the agreed attributes to interface with the database.

Otherwise, this'd break the backend and frontend services.

The complete response format the backend is expecting should look like this:
```
Article {
 id: ID <- A String
 number: Int
 content: String
 keywords: [String]
 wordCount: Int
}
```

services/parser-database/connector.py
```diff
@@ -73,8 +101,10 @@ def save_keywords_in_memory(keywords, article):
...
-            "frequency": frequency
+            "frequency": frequency,
+            "wordCount": article["wordCount"]
```
jaimehisao, line 107: Why are you adding an ID to the keywords and a word count? This information is already available elsewhere.

services/parser-database/connector.py
jaimehisao: Please don't break my code.

Domene99: Resolved as per oscar's private request

services/parser-database/connector.py
Domene99: Where is this info available? Since I don't know in depth your code this was what I managed to figure out

services/parser-database/connector.py
oscardavidrm: I need this @jaimehisao HAHAHAH.

Gimme my attributes!!!!!

services/parser-database/connector.py
```diff
@@ -61,6 +62,33 @@ def get_articles_that_match_keywords(keywords_list):
...
+            for article in keywords_in_memory[keyword]:
+                # tfidf computation
+                term_density_in_article = article["frequency"]/article["wordCount"]
+                document_frequency = len(articles_in_memory)/len(keywords_in_memory[keyword])
+                inverse_doc_freq = np.log(document_frequency)
+                value = term_density_in_article * inverse_doc_freq
+
+                articles_that_match_keyword[str(article["id"])] = value
+        matching_articles[keyword] = articles_that_match_keyword
+    return matching_articles
```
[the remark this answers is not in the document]
oscardavidrm: +1

jaimehisao:
I don't agree with Oscar on giving a free unit testing pass. For example, my tests can catch any errors on my code, but if Javier's code gets broken for any reason, there won't be a catch for that.

services/parser-database/connector.py
jaimehisao: The article dictionary

services/parser-database/tests/test_connector.py
```diff
@@ -21,18 +22,47 @@
...
+        {"id": 45, "frequency": 1},
+        {"id": 99, "frequency": 7},
     ],
     "licencia": [
-        {"id": "89", "articleNumber": 89, "frequency": 3},
-        {"id": "45", "articleNumber": 45, "frequency": 3},
-        {"id": "125", "articleNumber": 125, "frequency": 2},
+        {"id": 89, "frequency": 3},
+        {"id": 45, "frequency": 3},
+        {"id": 125, "frequency": 2},
```
jaimehisao, line 32: Why is removing `articleNumber` necessary?

Domene99: this is the keywords_in_memory mock and keywords_in_memory no longer has that attribute as per your newest change.

services/parser-database/connector.py
```diff
@@ -3,6 +3,7 @@
 # import requests  # pylint: disable=import-error
 import random
 import logging
+import numpy as np
```
anniefu, line 6: Does this need to be added to requirements.txt or is this a python standard library?

services/parser-database/connector.py
```diff
@@ -73,7 +102,7 @@ def save_keywords_in_memory(keywords, article):
         if keyword not in keywords_in_memory:
             keywords_in_memory[keyword] = []
         keywords_in_memory[keyword].append({
-            "articleNumber": article["articleNumber"],
+            "id": article["articleNumber"],
```
anniefu, line 105: Won't this cause a change in the response for either the `articles/byKeyword`? Was this a bug before or does all the callers of that endpoint need to change too?

services/parser-database/connector.py
jaimehisao: I think it needs to be added

Domene99:
> Nice find! Curious to know if you saw an improvement in the results you saw vs. the golden set queries you have.

services/parser-database/connector.py
jaimehisao: This was a bug that was fixed with another PR that was submitted.

services/parser-database/tests/test_connector.py
```diff
@@ -35,6 +35,35 @@
...
+
+articles_in_memory_no_wordCount = {'5': {}, '45': {}, '89': {}, '99': {}, '125': {}}
+
+
+def logn(num):
+    """
+    Mocks the natural log of a number to try to
+    minimize decimal points
+    """
+    return num
```
ssmall, line 64: nit: this is very questionable, but I'll allow it 😆

services/parser-database/tests/test_connector.py
```diff
@@ -71,6 +100,39 @@ def test_get_articles_that_match_keywords_non_empty_result_two_keywords():
...
+        assert expected == connector.get_articles_by_tfidf_value(keywords)
+
+
+@mock.patch("connector.keywords_in_memory", in_memory_value_mock_no_decimals)
+@mock.patch("connector.articles_in_memory", articles_in_memory_no_wordCount)
+def test_get_articles_by_tfidf_value_missing_word_count():
+    keywords = ["licencia", "ciclista"]
+    with mock.patch("numpy.log", side_effect=logn):
+        with pytest.raises(KeyError):
+            connector.get_articles_by_tfidf_value(keywords)
```
ssmall, line 133: Should. this really be an error? I'm looking through the code and trying to understand why this is a valid behavior. Under what circumstances do you want to return an error rather than an empty result?

services/parser-database/tests/test_connector.py
Domene99: hahhaha yeah it was my only idea on how to minimize decimal points

services/parser-database/tests/test_connector.py
Domene99: Well here I'm just saying that if there's no word_count attribute I cannot calculate the tfidf value. I think an error here is more valuable than an empty result for debugging purposes.

services/parser-database/tests/test_connector.py
Domene99: maybe int(log(num)) might be a little less questionable

services/parser-database/tests/test_connector.py
ssmall: Ok, I see now. Thanks for explaining.

PR #55, opened 2020-08-20
Fixing Key value that causes errors

jaimehisao:
Closing #57


Domene99:
> hmm, I would have expected a test case to change/be added that simulates the error to make sure the fix works & doesn't accidentally get added back or reverted by someone.

The current tests were not passing, so they already test for this bug. More info is on the issue page for this bug.

Domene99:
> Smh, who merged the broken test?
> 
> (but actually I have also merged broken PRs into master of repos that didn't have proper automated tests, and maybe once or twice to a repo that did)

HAHAHAH it was me, I think somewhere I accidentally deleted the type casting

PR #56, opened 2020-08-20
feature (cache) Add caching service to database requests.

### LINKED ISSUE #58: Setup caching for backend service

oscardavidrm:
+ Added a Redis server instance to cache articles whenever they get requested from the database service.
+ Added `async-redis` package.
+ A local Redis instance is necessary to take advantage of the caching.

*** Notes
Two new .env variables where added `REDIS_IP` and `REDIS_PORT`. Both should be added to the Cloud Run container for the backend service to make them available.

Closes #58


server/Dockerfile
```diff
@@ -1,8 +1,10 @@
 FROM node:10
 
 ENV PORT=4000
-ENV SEARCH_ENDPOINT="http://localhost:8080"
-ENV DATABASE_ENDPOINT="http://localhost:8082"
+ENV SEARCH_ENDPOINT="http://host.docker.internal:8080"
```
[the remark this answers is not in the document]
oscardavidrm: Those ones get served as env variables inside the Cloud Run container for the backend service. This is best practice as it allows to have separate local/production environments.

server/index.js
```diff
@@ -1,19 +1,22 @@
...
 import { importSchema } from 'graphql-import'
 import path from 'path'
+import redis from 'async-redis'
 import resolvers from './resolvers'
-import { PORT } from './config'
+import { PORT, REDIS_IP, REDIS_PORT } from './config'
 
 const typeDefs = importSchema(path.join(__dirname, './typedefs/index.graphql'))
 
+const cache = redis.createClient({ host: REDIS_IP, port: REDIS_PORT })
```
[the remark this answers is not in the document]
oscardavidrm: According to the documentation, there shouldn't be any problem joining a Redis server from within any other component inside our GCP VPC. So I would not expect any issues, but that's left to see.

server/Dockerfile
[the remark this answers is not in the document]
oscardavidrm: Plus I don't think it's a good idea to expose the database endpoint in Github 😆 . The bill would be too high trying to survive a DoS attack.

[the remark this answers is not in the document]
Domene99: whooops, @jaimehisao and I have exposed our endpoints

server/index.js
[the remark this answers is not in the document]
Domene99: Is the redis instance a gcp service? Let's hope there's no issue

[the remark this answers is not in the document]
oscardavidrm: Yes, exactly.

Is a service offered by GCP, and it exposes an IP address as well.

server/resolvers/article/queries.js
```diff
@@ -2,23 +2,38 @@ import { AVERAGE_WORDS_PER_MINUTE } from '../../utils/constants'
...
     const payload = await searchApi(search)
 
     if (payload.error) return new Error(JSON.stringify(payload.error))
 
     const articleIds = [...payload.data.articles]
     const articles = []
+    const requests = []
+
+    for (const id of articleIds) {
+      if (await cache.exists(id)) {
```
Domene99, line 15: What happens if the id in the cache exists but the road regulations document has been updated. Should we drop the cache at some point?

oscardavidrm: Yes, that would be a nice stretch for the parser service.

Whenever it re-parses the document in-memory, it should drop the cache content so it can start fresh and store the updated data.

server/resolvers/article/queries.js
```diff
@@ -2,23 +2,38 @@ import { AVERAGE_WORDS_PER_MINUTE } from '../../utils/constants'
 import { databaseApi, searchApi } from '../../endpoints'
 
 const articleQueries = {
-  articles: async (_, { search }) => {
+  articles: async (_, { search }, { cache }) => {
```
anniefu, line 5: You'll probably want a test that mocks this cache out with an in-memory cache just for testing. Then you can test what's in the in-memory cache at the end of the test to make sure that things are being added correctly.

oscardavidrm: Oh okay, so an object to mock the cache should do it right? @anniefu 

Something like `const cache = new Map()` which I can use as a local cache. The cool thing about this DS in Javascript is that it supports ops such as `set` and `get` which are essentially the same that **redis** uses to store `key/value` pairs.

server/tests/queries.test.js
```diff
@@ -9,8 +9,16 @@ import resolvers from '../resolvers'
 import { databaseApi, searchApi } from '../endpoints'
 import { mockArticles } from './mocks'
 
+// In-memory cache to mock redis operations
```
oscardavidrm, line 12: Wdyt? @anniefu

anniefu: Nice! JavaScript is pretty wild with this ability to just overwrite a method on an object.

server/index.js
```diff
@@ -2,22 +2,25 @@ import { ApolloServer } from 'apollo-server-express'
...
+
 const server = new ApolloServer({
   typeDefs,
   resolvers,
-  context: ({ req = {}, res }) => ({ req, res }),
+  context: ({ req = {}, res }) => ({ req, res, cache }),
```
ssmall, line 16: nit: consider defaulting to no cache when the `REDIS_IP` is not specified. This way, a developer can still test locally without having to spin up their own Redis instance.

server/index.js
```diff
@@ -2,22 +2,25 @@ import { ApolloServer } from 'apollo-server-express'
...
   introspection: true,
   playground: true
 })
 
+cache.on('ready', () => console.log(`🧶  Redis server ready at ${REDIS_IP}:${REDIS_PORT}`))
```
ssmall, line 21: Do you need to wait for the cache to be ready before starting the app? What if it is never ready (due to a bad IP address, for example)?

server/tests/queries.test.js
```diff
@@ -116,6 +124,13 @@ describe('Queries', () => {
       if (results.errors) {
         return expect(results.errors).toEqual(expected)
       }
+
+      // Format resulting article as how they should be stored in the cache
+      const cacheToExpect = new Map(results.data.articles.map(article => ([article.id, JSON.stringify({ ...article })])))
+
+      expect(cache).toEqual(cacheToExpect)
+      cache.clear()
```
ssmall, line 132: can you also add a test where the cache is already "warm" (already has some articles in it) and assert that the database api does not get called in that case?

ssmall: It doesn't look like this was addressed. Somewhere in the test there needs to be an assertion like

```javascript
expect(databaseApi.mock.calls).toContainEqual(...)
```

[Reference](https://jestjs.io/docs/en/mock-functions#custom-matchers)

PR #65, opened 2020-08-20
Add Python Unit Testing

jaimehisao:
Python Unit testing for all the things!

### DIFF
--- a/.github/workflows/python-unit-testing.yml
+++ b/.github/workflows/python-unit-testing.yml
@@ -1,0 +1,53 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python Services Unit Tests
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies (Keywords)
+      working-directory: ./services/keywords
+      run: |
+        python -m pip install --upgrade pip
+        pip install pytest
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        
+    - name: Install dependencies (Search Engine)
+      working-directory: ./services/search/
+      run: |
+        python -m pip install --upgrade pip
+        pip install pytest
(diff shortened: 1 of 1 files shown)


.github/workflows/python-unit-testing.yml
```diff
@@ -0,0 +1,53 @@
...
+      working-directory: ./services/parser-database
+      run: |
+        python -m pip install --upgrade pip
+        pip install pytest pytest-mock
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+  
+    - name: Test with pytest (parser)
+      working-directory: ./services/parser-database
+      run: |
+        pytest
```
anniefu, line 45: Any reason for `pytest` here but `python -m pytest` for the others?

.github/workflows/python-unit-testing.yml
```diff
@@ -0,0 +1,53 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies (Keywords)
+      working-directory: ./services/keywords
```
anniefu, line 22: Maybe just make a "Install pytest" step as the first step so you don't have to repeat?

jaimehisao: Mostly having trouble because every separate app has a different requirements.txt and I need to pip install everything inside them

.github/workflows/python-unit-testing.yml
jaimehisao: No, changed it to look the same

.github/workflows/python-unit-testing.yml
anniefu: Yeah, that's ok. You could write a small script to find all the requirements.txt and install them, but this is fine.

I would move your `pip install pytest pytest-mock` to its own step at the beginning. Otherwise, if someone switches the order of installing search engine dependencies and keywords dependencies, the unit tests will confusingly break.

.github/workflows/python-unit-testing.yml
```diff
@@ -0,0 +1,49 @@
...
+      run: |
+        python -m pytest
+    - name: Test with pytest (keyword)
+      working-directory: ./services/keywords
+      run: |
+        python -m pytest
+    - name: Test with pytest (searchengine)
+      working-directory: ./services/search
+      run: |
+        python -m pytest
```
ssmall, line 49: This actually seems like a great use case for a [matrix strategy](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategy). That way you can define a generic job template that will get executed for each service.

PR #67, opened 2020-08-20
quick fix: change utils to synonym_extractor

Domene99:
Closing #66


Domene99:
> Hmm, none of the unit tests run across L62? Or there was one but it was failing?

The unit tests run inside synonym_extraction

anniefu:
But the unit tests in here should have broken, right? https://github.com/googleinterns/major-tom/blob/main/services/search/testing/test_search_engine.py

Otherwise, you'll probably want to add a test that would have caught this by testing the SearchEngine constructor.

Domene99:
Turned out to be a quick fix with a big test to back it up

services/search/testing/test_search_engine.py
```diff
@@ -1,11 +1,25 @@
 from unittest import mock
 import json
+import flask  # pylint: disable=import-error
 import responses  # pylint: disable=import-error
 from search_engine import SearchEngine  # pylint: disable=import-error
 import test_constants as constants
+from main import search_service  # pylint: disable=no-name-in-module
```
ssmall, line 7: if you just did `import main` would you be able to avoid the pylint error?

PR #75, opened 2020-08-22
refactor (match api spec) Update left non-existent attribute.

oscardavidrm:
+ The previous refactor aimed at updating the following attributes:
 `articleNumber` -> `number`
 `text` -> `content`
  Added `id`

+ Updated missing attribute in `parser-database/connector.py`.


services/parser-database/connector.py
```diff
@@ -74,7 +74,7 @@ def save_keywords_in_memory(keywords, article):
             keywords_in_memory[keyword] = []
         keywords_in_memory[keyword].append({
             "number": article["number"],
-            "id": article["articleNumber"],
+            "id": article["id"],
             "frequency": frequency
```
anniefu, line 78: Shouldn't this have broken some tests?

oscardavidrm: Just ran them over again and everything passed 👍

ssmall: Was there a failing test before you changed this? If not, there is a unit test missing and should be added in this PR to avoid a similar breakage in the future.

anniefu: Hmm, I assume the unit tests are mocking this out correctly, but the actual service was not returning "id". As long as the unit tests are using "id" also instead of "articleNumber", I'm fine with this.

ssmall: I don't think anything is being mocked to test this function -- it should be a pretty simple test and an obvious failure

oscardavidrm: Okay, let me add the test

Domene99:
After digging through the code I found that the untested functions are save_keywords_in_memory, and store_articles. Everything passed in the previous run because keywords_in_memory was being mocked with the correct keys when save_keywords_in_memory used the incorrect ones.

Domene99:
I think a test that asserts no errors for both of this functions should work

PR #78, opened 2020-08-24
Add test for save keywords in memory

Domene99:
asserts for no exceptions and asserts correct saved keywords
Closing #79

### DIFF
--- a/services/parser-database/tests/test_connector.py
+++ b/services/parser-database/tests/test_connector.py
@@ -99,6 +99,7 @@ def test_keywords_in_memory():
                           "bicicleta": [{"id": 1, "number": 1, "frequency": 1}],
                           "usar": [{"id": 1, "number": 1, "frequency": 1}],
                           "casco": [{"id": 1, "frequency": 1, "number": 1}]}
+    assert {} == connector.keywords_in_memory
     try:
         connector.save_keywords_in_memory(
             ["forzoso", "bicicleta", "usar", "casco"],
--- a/services/parser-database/tests/test_connector.py
+++ b/services/parser-database/tests/test_connector.py
@@ -35,6 +35,18 @@ in_memory_value_mock = {
 }
 
 
+keywords_mock = ["forzoso", "bicicleta", "usar", "casco"]
+
+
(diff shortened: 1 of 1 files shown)


services/parser-database/tests/test_connector.py
```diff
@@ -92,3 +104,11 @@ def test_if_got_error_from_keywords_service(mock_get):
     mock_get.return_value.json.return_value = {'error': {'message': "something happened"}}  # noqa: E501
     with pytest.raises(Exception):
         assert connector.get_keywords(text_to_keywordize)
+
+
```
anniefu, line 108: [nit] I think this is more just `test_save_keywords_in_memory`

services/parser-database/tests/test_connector.py
```diff
@@ -35,6 +35,18 @@
 }
 
 
+keywords_mock = ["forzoso", "bicicleta", "usar", "casco"]
```
anniefu, line 38: These can be just local variables to the one test case. 

Also [nit] typically mocks refer to when you fake out the return of a service or function because you can't call the actual function, but in this case we are not circumventing anything, so you don't have to include `mock` in the name. These are just normal, valid inputs to this function.

services/parser-database/tests/test_connector.py
```diff
@@ -92,3 +92,17 @@ def test_if_got_error_from_keywords_service(mock_get):
...
+                          "bicicleta": [{"id": 1, "number": 1, "frequency": 1}],
+                          "usar": [{"id": 1, "number": 1, "frequency": 1}],
+                          "casco": [{"id": 1, "frequency": 1, "number": 1}]}
+    try:
+        connector.save_keywords_in_memory(
+            ["forzoso", "bicicleta", "usar", "casco"],
+            {"id": 1, "number": 1, "content": "en bicicleta es siempre forzoso usar casco"})
+        assert expected_in_memory == connector.keywords_in_memory
+    except Exception as e:
+        pytest.fail("Error: " + str(e))
```
ssmall, line 108: Good test! I only have one concern: because `connector.keywords_in_memory` is global state, it could be affected by other tests. The safest way to avoid this is to add
```
assert {} == connector.keywords_in_memory
```

at the *beginning* of the unit test. When you do that, does the test still pass?

PR #86, opened 2020-08-25
Create docker-image.yml

jaimehisao:


.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,42 @@
...
+      run: |
+        docker build . --file Dockerfile --tag search
+        docker tag search docker.pkg.github.com/googleinterns/major-tom/search
+        docker push docker.pkg.github.com/googleinterns/major-tom/search
+    - name: Build Parser Docker image
+      working-directory: ./services/parser-database
+      run: |
+        docker build . --file Dockerfile --tag parser-database
+        docker tag parser-database docker.pkg.github.com/googleinterns/major-tom/parser-database
+        docker push docker.pkg.github.com/googleinterns/major-tom/parser-database
```
ssmall, line 42: the `docker login` could be moved to a separate step and then the build step could be parameterized as a matrix job

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,42 @@
...
+    - name: Build Search Docker image
+      working-directory: ./services/search
+      run: |
+        docker build . --file Dockerfile --tag search
+        docker tag search docker.pkg.github.com/googleinterns/major-tom/search
+        docker push docker.pkg.github.com/googleinterns/major-tom/search
+    - name: Build Parser Docker image
+      working-directory: ./services/parser-database
+      run: |
+        docker build . --file Dockerfile --tag parser-database
```
oscardavidrm, line 40: I am pretty confident you can omit the `--file` flag as Docker looks for that file by default.

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,42 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build Frontend Docker image
+      working-directory: ./app
+      run: |
+        docker login docker.pkg.github.com -u ${{ secrets.USERNAME }} --password ${{ secrets.GITHUB_TOKEN }}
+        docker build . --file Dockerfile --tag frontend
+        docker tag frontend docker.pkg.github.com/googleinterns/major-tom/frontend
```
oscardavidrm, line 17: Correct me if I am wrong but, are these URL's correct? I thought that for pushing to Google's Container Registry you had to prefix your images with the following [format](https://cloud.google.com/container-registry/docs/pushing-and-pulling):
```
gcr.io/$PROJECT_ID/$IMAGE_NAME
i.e. gcr.io/major-tom-123/frontend
```

Domene99: I think this pr is about publishing our latest image to the repo not to gcp's container registry

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,42 @@
+name: Build Docker Images & Push to Registry
+
```
anniefu, line 2: I assume this is a cloudbuild yaml? I would leave cloudbuild in the name somewhere if so

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,42 @@
...
+      run: |
+        docker login docker.pkg.github.com -u ${{ secrets.USERNAME }} --password ${{ secrets.GITHUB_TOKEN }}
+        docker build . --file Dockerfile --tag frontend
+        docker tag frontend docker.pkg.github.com/googleinterns/major-tom/frontend
+        docker push docker.pkg.github.com/googleinterns/major-tom/frontend
+    - name: Build Backend Docker image
+      working-directory: ./server
+      run: |
+        docker build . --file Dockerfile --tag backend
+        docker tag backend docker.pkg.github.com/googleinterns/major-tom/backend
```
anniefu, line 23: Isn't this tag lost if you don't push it with this tag too?

PR #87, opened 2020-08-25
bugfix (cache) Add expiration to cached articles to flush them periodically.

### LINKED ISSUE #85: Articles should not be cached indefinitely
There needs to be some sort of automatic expiration on the cache used for the frontend, otherwise Article content will never be refreshed.

oscardavidrm:
+ Added `.expire` to every article set to last at most a day on the cache before being flushed. [resolvers/article/index.js]

+ Had to switch Redis dependency from `async-redis` to `redis` since the first one hasn't been updated in a while and lacks expiration support. Plus `redis` library is the official one suggested on Redis documentation. [package.json, yarn.lock]

+ The `redis` dependency comes with the hassle of not supporting asynchronous calls out of the box, but, as per their documentation, in order to force it to be asynchronous, the built-in `util` library from Node.js allows us to wrap the Redis ops in asynchronous calls. [index.js]

Closes #85

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1,8 +1,9 @@
 import { ApolloServer } from 'apollo-server-express'
 import { importSchema } from 'graphql-import'
+import { promisify } from 'util'
 import express from 'express'
 import path from 'path'
-import redis from 'async-redis'
+import redis from 'redis'
 import localCache from './utils/in-memory-cache'
 import resolvers from './resolvers'
 import { PORT, REDIS_IP, REDIS_PORT } from './config'
@@ -11,9 +12,16 @@ const typeDefs = importSchema(path.join(__dirname, './typedefs/index.graphql'))
 
 // If you don't want to spin up a local Redis server, make sure to avoid
 // REDIS_IP and REDIS_PORT .env variables declared to default to an in-memory storage
-const cache = REDIS_IP && REDIS_PORT
-  ? redis.createClient({ host: REDIS_IP, port: REDIS_PORT })
-  : localCache
+let cache
+if (REDIS_IP && REDIS_PORT) {
+  cache = redis.createClient({ host: REDIS_IP, port: REDIS_PORT })
+
+  // Make Redis ops asynchronous.
+  cache.get = promisify(cache.get).bind(cache)
+  cache.exists = promisify(cache.exists).bind(cache)
+} else {
+  cache = localCache
+}
 
 const server = new ApolloServer({
   typeDefs,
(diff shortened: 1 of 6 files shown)


server/resolvers/article/queries.js
```diff
@@ -32,6 +32,7 @@ const articleQueries = {
       delete article.wordCount
 
       cache.set(article.id, JSON.stringify(article))
+      cache.expire(article.id, 60 * 60 * 24)
```
jaimehisao, line 35: great!

server/index.js
```diff
@@ -1,8 +1,9 @@
 import { ApolloServer } from 'apollo-server-express'
 import { importSchema } from 'graphql-import'
+import { promisify } from 'util'
```
Domene99, line 3: What's the difference of using promisify vs async-redis?

oscardavidrm: If you noticed, I switched dependendencies, from `async-redis` to `redis`. The latter one is the official one suggested in the actual Redis docs. The  only drawback it has, it doesn't support yet `async/await` asynchronous calls out of the box.

`promisify` is a built-in Node.js dependency to wrap the `redis` operations in an async manner. Thus allowing us to do something like:
```
const article = await cache.get(id);
```
rather than:
```
cache.get(id, callback => {...})
```

So actually, `async-redis` its not being substituted by `promisify`, but rather by `redis`. `promisify` just allows us to have that "syntactic sugar" available for the cache.

server/resolvers/article/queries.js
ssmall, line 35: This method doesn't exist for the in-memory cache, right? What happens when you try to run this code with an in-memory cache?

oscardavidrm: True lol.

I can go ahead and simulate that behavior with a timeout for the local storage.
Something like:
```
localCache.expire = (key, ttl) => setTimeout(() => localCache.delete(key), ttl * 1000)
```

server/utils/in-memory-cache.js
```diff
@@ -1,6 +1,7 @@
 const localCache = new Map()
 
 localCache.exists = key => localCache.has(key)
+localCache.expire = (key, ttl) => setTimeout(() => localCache.delete(key), ttl * 1000)
```
ssmall, line 4: General shout-out that this is the kind of situation where Typescript is really valuable. If you were using Typescript, the missing `expire` function would have been caught as a compiler error rather than during code review.

Just food for thought, not something I expect you to change in this PR.

PR #88, opened 2020-08-25
enhancement (welcome-message) Add welcome message and fix display styles.

### LINKED ISSUE #81: Add a landing welcome message with explanation.
The current first interaction with the application does not really introduce the user as to what the application does.

Some sort of welcoming guide explaining what the application is about would be nice!

oscardavidrm:
+ Added a more descriptive placeholder for the search bar. [components/search-bar/index.js]
+ Modified CSS margin attributes for Search Bar to fit better the new results counter. [components/search-bar/styles.js]
+ Added a new Welcome message card that displays whenever there are no articles to display. The card introduces the user as to what the application does and an example to start using the service. [views/search-engine/index.js, views/search-engine/styles.js]

Closes #81

### DIFF
--- a/app/src/views/search-engine/index.js
+++ b/app/src/views/search-engine/index.js
@@ -1,11 +1,12 @@
 import React, { useState, useEffect } from 'react'
 import { useQuery } from '@apollo/client'
 import { useDebounce } from 'use-lodash-debounce'
-import { Container, Snackbar } from '@material-ui/core'
+import { Card, CardContent, Container, Snackbar, Chip, Typography } from '@material-ui/core'
 import { Alert } from '@material-ui/lab'
 import SearchBar from './components/search-bar'
 import ArticlesList from './components/articles-list'
 import { GET_ARTICLES } from './graphql/queries'
+import styles from './styles'
 
 const SearchEngine = () => {
   const [loading, setLoading] = useState(false)
@@ -14,6 +15,8 @@ const SearchEngine = () => {
   const [search, setSearch] = useState('')
   const debouncedSearch = useDebounce(search, 1000)
   const getArticlesQuery = useQuery(GET_ARTICLES, { variables: { search: debouncedSearch } })
+  const classes = styles()
+  const DOC_URL = 'https://www.guadalupe.gob.mx/nuevo-reglamento-de-transito-homologado/'
 
   useEffect(() => {
     const { data, error } = getArticlesQuery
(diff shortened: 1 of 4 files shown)


app/src/views/search-engine/index.js
```diff
@@ -44,7 +47,48 @@ const SearchEngine = () => {
...
+            <Typography gutterBottom variant='h5' component='h2'>
+                Motor de búsqueda - Reglamento Vial de Monterrey
+            </Typography>
+            <Typography
+              variant='body1'
+              color='textSecondary'
+              component='p'
+              className={classes.cardContent}
+            >
+                Algoritmo para realizar búsquedas optimizadas sobre los <b>reglamentos viales oficiales</b> de tu comunidad.
```
Domene99, line 68: I'd like an english card as well to make it easier for the hiring committe or any english only reader to understand

app/src/views/search-engine/index.js
```diff
@@ -44,7 +47,48 @@ const SearchEngine = () => {
         setLoading={setLoading}
         loading={loading}
       />
+      <Chip
+        size='small'
+        label={`${articles.length} resultados`}
+        className={classes.resultsChip}
+      />
       <ArticlesList articles={articles} />
+      {articles.length === 0 &&
```
Domene99, line 56: If a query returns no articles the info card will be displayed?

app/src/views/search-engine/index.js
ssmall: I wouldn't worry too much about this. For users who don't speak Spanish, Chrome will automatically offer to translate the page for them.

app/src/views/search-engine/index.js
```diff
@@ -14,6 +15,8 @@ const SearchEngine = () => {
   const [search, setSearch] = useState('')
   const debouncedSearch = useDebounce(search, 1000)
   const getArticlesQuery = useQuery(GET_ARTICLES, { variables: { search: debouncedSearch } })
+  const classes = styles()
+  const DOC_URL = 'https://www.guadalupe.gob.mx/nuevo-reglamento-de-transito-homologado/'
```
ssmall, line 19: This is okay as a short term fix, however in the long term you and @jaimehisao should figure out how this value could be determined dynamically. Once you add more jurisdictions, it does not make sense to hardcode this URL.

app/src/views/search-engine/index.js
oscardavidrm: Yes, that's correct!

app/src/views/search-engine/index.js
oscardavidrm: Totally agree. Once having a DB where we store the documents, fetching its information is what we should be doing.
Something like:

```
const document = await fetch('/document/id');
document.URL
document.name
...
```

That way we'd achieve a more dynamic frontend. But yeah, that's something to look into.

app/src/views/search-engine/index.js
Domene99: I forgot to respond to this, but won't this be super confusing for the user

anniefu: +1 it'd be nice to display a no results found or something so that users don't think your app is just taking forever

app/src/views/search-engine/index.js
anniefu: I actually don't think it'd be a bad idea unless it'd take a long time. Or at least to make an hideable English version that users can click on. A good portion of your audience, including the people you present to in the final STEP presentation will be primarily English speaking. 

If you choose not to include an English card, you'll want to double check all the text translates well with Google translate.

(Also, what if someone uses not-Chrome? *horrified gasp*)

app/src/views/search-engine/index.js
oscardavidrm: I actually added a small results count right below the search bar!! So yeah, you actually get some feedback in the results count.

PR #89, opened 2020-08-26
bugfix (article card) Set maxHeight on article cards to fix height issue.

### LINKED ISSUE #84: Preview height for article cards not constant
Preview height for article cards is now really inconsistent between article cards making some really long, I'm guessing this is due to the new unicode formatting.

oscardavidrm:
+ Added a `maxHeight` attribute to the card content to limit the size of all cards.

Closes #84

### DIFF
--- a/app/src/views/search-engine/components/articles-list/styles.js
+++ b/app/src/views/search-engine/components/articles-list/styles.js
@@ -9,7 +9,8 @@ const styles = makeStyles(() => ({
   },
   cardContent: {
     whiteSpace: 'pre-wrap',
-    marginBottom: 15
+    marginBottom: 15,
+    maxHeight: '20vh'
   }
 }))
 


app/src/views/search-engine/components/articles-list/styles.js
```diff
@@ -9,7 +9,8 @@ const styles = makeStyles(() => ({
   },
   cardContent: {
     whiteSpace: 'pre-wrap',
-    marginBottom: 15
+    marginBottom: 15,
+    maxHeight: '20vh'
```
ssmall, line 13: TIL about `vh` as a CSS unit. vcool!

oscardavidrm: Hahahaha yeah!! There are a bunch of [units in CSS](https://www.w3schools.com/cssref/css_units.asp).

PR #90, opened 2020-08-26
Add golden set

Domene99:
Closing #44


golden-set.md
```diff
@@ -0,0 +1,93 @@
...
+## Query: La carroceria
+## Output: 4, 24,  166
+
+* Carroceria
+  * 4 - 1
+  * 24 - 1
+  * 166 - 2
+
+## Query: Permiso de conducir
+## Output: 4, 24, 32, 43, 49, 51, 107, 108, 109, 110, 111, 112, 115, 117, 124, 126, 130, 131, 134, 166, 170, 171, 173, 175
```
ssmall, line 65: Is this output ordered by relevance? If so, I'm curious why 166 is not the first result.

golden-set.md
```diff
@@ -0,0 +1,93 @@
...
+  * None
+
+## Query: Choque
+## Output: 137, 166, Anexo 5
+* 137 - 5
+* 166 - 5
+* Anexo 5 - 1
+
+## Query: ¿Me pueden multar por un neumático ponchado?
+## Output: 24, 130, 136, 158, 166, 168, 169, 172, 173
```
ssmall, line 43: Why is 166 not higher in the list given that it has the most matches?

golden-set.md
Domene99: I think this question is for @oscardavidrm but it seems he ordered them ascendingly

oscardavidrm: Yes, I just wrote them as I found them. There's no sorting in there.

PR #91, opened 2020-08-26
Re raise status codes which are not 200

Domene99:
Closing #77


Domene99:
> I just did some research on `raise_for_status()` op and how you're implementing it seems reasonable!
> 
> Nice!

And it's really cool because of the way we are managing errors, they get re-raised to the main function of the endpoint and processed there without the need of any more try catch blocks.

services/search/search_engine.py
```diff
@@ -41,6 +41,7 @@ def query(self, query):
         logging.debug("keywords location: %s", env.get_keyword_endpoint())
 
         response = requests.post(env.get_keyword_endpoint(), json=query_text)
+        response.raise_for_status()
         response = response.json()
```
anniefu, line 45: [nit] technically unrelated to this PR, but overwriting a variable with a different object type (`.json()` returns a string) is a weird python thing, and probably bad practice as it can lead to confusion to people reading the code who only noticed one of the assignments to `response`

Domene99: I just assumed my reviewers are perfect beings

services/parser-database/requirements.txt
```diff
@@ -3,4 +3,5 @@ utils
 https://github.com/timClicks/slate/archive/master.zip
 requests
 numpy
-pytest-mock
\ No newline at end of file
+pytest-mock
```
jaimehisao, line 6: Does this need to be in the production image?

PR #92, opened 2020-08-28
Database implementation

jaimehisao:
Started implementing the database without unit testing. While probably it won't get merged but here is the code.


services/parser-database/app.py
```diff
@@ -50,5 +50,5 @@ def get_article_by_number_in_memory(id):
 
 
 if __name__ == '__main__':
-    parse_all_documents()
+    #parse_all_documents()
```
Domene99, line 53: This shouldn't be commented should it?

services/parser-database/connector.py
```diff
@@ -2,15 +2,56 @@
...
+    @staticmethod
+    def from_dict(src):
+        self.number = src["number"]
+        self.id = src["id"]
+        self.content = src["keyword"]
+        self.wordCount = src["frequency"]
+
+cred = credentials.ApplicationDefault()
+firebase_admin.initialize_app(cred, {
+  'projectId': 'major-tom-285619',
```
Domene99, line 37: Is this needed, wouldn't the cloud run containers already run on this projectId and be defaulted on initialize_app?

services/parser-database/connector.py
```diff
@@ -2,15 +2,56 @@
...
+
 articles_in_memory = {}
 keywords_in_memory = {}
 
 
+def get_documents_to_parse_db():
+    documents_ref = db.collection(u'documents')
+    docs = documents_ref.stream()
+
+    for doc in docs:
```
Domene99, line 50: this seems like a debugging loop, either remove or change to log

services/parser-database/connector.py
```diff
@@ -83,3 +152,56 @@ def store_article(article_dict):
...
+    return matching_articles
+
+
+def get_article_by_id_db(art_num):
+    documents_ref = db.collection(u'articles').document(art_num)
+    doc = documents_ref.get()
+    if doc is not None:
+        return doc.to_dict()
+    else:
+        return None
```
Domene99, line 207: Maybe we should return some info for debugging purposes

services/parser-database/connector.py
```diff
@@ -61,6 +102,34 @@ def get_articles_that_match_keywords(keywords_list):
     return matching_articles
 
 
+def get_articles_by_tfidf_value(keywords_list):
```
Domene99, line 105: This is using the in_memory storage which needs to be changed to db storage just like get_articles_that_match_keywords

services/parser-database/connector.py
```diff
@@ -83,3 +152,56 @@ def store_article(article_dict):
...
+        frequency = article["content"].count(keyword)
+
+        doc_ref = db.collection(u'keywords').where('keyword', '==', keyword)
+        doc = doc_ref.get()
+        
+        if len(doc) != 0 and doc[0] is not None:
+            from_db = doc[0].to_dict()
+            print(from_db)
+            from_db["matching_articles"][article["id"]] = frequency
+            #print(from_db)
```
Domene99, line 180: Some leftover prints

jaimehisao:
> Overall pretty impressive, I can't pinpoint where the code is becoming so slow, from 1 to 10 mins is more than just communication delay

I don´t know either, but it is probably not on my end since it is the same code as the in-memory implementation.

services/parser-database/app.py
jaimehisao: It is left out so that every-time the app starts it doesn't have to parse the doc and I can test quickly.

services/parser-database/connector.py
jaimehisao: If it is not there I cannot test locally

services/parser-database/connector.py
jaimehisao: Still in implementation phase.

jaimehisao:
Would like to comment that the PR is still in draft form so there might be a lot of prints there and comments too, those are being addressed.

services/parser-database/connector.py
jaimehisao: Something like your last PR where you added more error handling?

services/parser-database/app.py
```diff
@@ -7,8 +7,8 @@
 from flask import jsonify  # pylint: disable=import-error
 
 from parser import parse_all_documents
-from connector import get_articles_that_match_keywords
-from connector import get_article_by_number
+from connector import get_article_by_id_db
+from connector import get_articles_that_match_keywords_db
```
ssmall, line 11: As @anniefu mentioned, you'll probably want to just have a `db_connector.py` that defines methods with the same names that you already use in the code, so that you can just change the import statement and not have to change any code.

services/parser-database/connector.py
ssmall: So this will be removed before merging?

services/parser-database/connector.py
```diff
@@ -83,3 +152,56 @@ def store_article(article_dict):
...
+            db.collection(u'keywords').document(doc[0].id).set(from_db)
+        else:
+            to_send = {"keyword": keyword, "matching_articles": {article["id"]: frequency}}
+            db.collection(u'keywords').add(to_send)
```
ssmall, line 184: It looks like you're trying to mirror the same structure that you had for the in-memory implementation, which is not necessarily the best implementation. For example, `keywords` could be a field of the article document that gets queried directly, which eliminates the need to store `keywords` as a separate collection. See [this guide](https://firebase.google.com/docs/firestore/query-data/queries) on how to perform different queries against Firestore, particularly the example for [array membership](https://firebase.google.com/docs/firestore/query-data/queries#array_membership).

services/parser-database/parser.py
```diff
@@ -38,7 +38,7 @@ class Article:
     def __init__(self, number, content):
         self.number = number
         self.content = content
-        self.id = str(number)
+        self.id = 'monterrey'+str(number)
```
ssmall, line 41: Could this ID be generated by Firebase instead?

